PR #344, opened 2023-01-18
Remove unsupported immutable privileges management option

l-heemann:
Security init file is not a supported option and should not be mentioned in the docs

Deleted documentation has moved to this internal doc: https://sites.google.com/neotechnology.com/dbms-internal-documentation/administering-immutable-privileges


renetapopova:
I guess this should also go into 5.x. So, I'll do that.

PR #416, opened 2023-01-30
put advanced clustering back, only on dev

AlexicaWright:


AlexicaWright:
Incorrect label, but is fine now, no need to cherry-pick.

PR #608, opened 2023-03-23
Moving `sub` section to the top 

lidiazuin:
As per request, this is not an Azure-only setting, but should be applied in any case.


lidiazuin:
Closing this PR which was re-created here https://github.com/neo4j/docs-operations/pull/625 after the detachment of forks.

PR #632, opened 2023-03-31
Fix the table Available commands per category

renetapopova:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-632.surge.sh

PR #652, opened 2023-04-11
Add a warning that neo4j-admin database upload does not support private linking

renetapopova:

PR #657, opened 2023-04-11
Add a warning that neo4j-admin push-to-cloud does not support private linking

renetapopova:
Copied from #652


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-657.surge.sh

PR #673, opened 2023-04-14
Updating a last command that was still install-service

lidiazuin:
Instead of the correct `windows-service`

PR #675, opened 2023-04-14
Updating a last command that was still install-service (#673)

lidiazuin:
Instead of the correct `windows-service`

Cherry-picked from https://github.com/neo4j/docs-operations/pull/673


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-675.surge.sh

PR #674, opened 2023-04-14
Better document the username and groups config items

mnd999:
The groups config item gained support for a single string in 5.4.0 so we should reflect that. Also explain in more detail what is supported.

PR #678, opened 2023-04-14
Better document the username and groups config items (#674)

renetapopova:
The groups config item gained support for a single string in 5.4.0 so we should reflect that. Also explain in more detail what is supported.

Cherry-picked from #674


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-678.surge.sh

PR #700, opened 2023-04-18
updated debian install instructions for if openjdk 19 gets installed

jennyowen:
Also added RHEL 8.8 to compatible OSs.

### DIFF
--- a/modules/ROOT/pages/installation/linux/debian.adoc
+++ b/modules/ROOT/pages/installation/linux/debian.adoc
@@ -11,6 +11,7 @@ Neo4j {neo4j-version} requires the Java 17 runtime.
 Java 17 is not included in Ubuntu 16.04 LTS and will have to be set up manually prior to installing or upgrading to Neo4j {neo4j-version}, as described below.
 Debian 11 and Ubuntu 18.04 onwards already have the OpenJDK Java 17 package available through `apt`.
 
+[[debian-prerequisites-notopenjdk]]
 === Oracle JDK, Zulu JDK, or Corretto JDK
 
 If you wish to use a non-default JDK, it must be installed prior to starting the Neo4j installation.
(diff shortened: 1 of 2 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-700.surge.sh

modules/ROOT/pages/installation/linux/debian.adoc
```diff
@@ -55,21 +55,58 @@ java-1.17.0-openjdk-amd64 1711 /usr/lib/jvm/java-1.17.0-openjdk-amd64
...
+
+If this happens, you will see this warning on Neo4j start:
+[output]
+----
+WARNING! You are using an unsupported Java runtime.
+* Please use Oracle(R) Java(TM) 17, OpenJDK(TM) 17 to run Neo4j.
+* Please see https://neo4j.com/docs/ for Neo4j installation instructions.
+----
+
+To fix this, you can install Java 17 manually, uninstall OpenJDK 19, or set Java 17 as the default.
```
jennyowen, line 89: Sadly no. If you uninstall java 19 then apt will helpfully also uninstall neo4j UNLESS there's a compatible java version already installed.
So basically what happens is that:
1. neo4j says it has a dependency on java-17. 
2. OpenJDK 19 package says that it provides a compatible java-17 package. 
3. Java 19 is a more recent version than java 17, so apt assumes that it's better.
4. apt """helpfully""" installs the more recent java 19 package.
5. we come along and uninstall java 19 because we don't want it.
6. nothing provides a java 17 any more, so neo4j has a missing dependency.
7. apt also deletes neo4j because it won't be able to run without a java 17.

To stop apt deleting neo4j in step 6, we need to have java 17 installed already so that there is a package installed which provides java-17.

I suppose there are really two fix options:
1. install java 17 and uninstall java 19
2. install java 17 and set it to default so that java 19 is ignored.

PR #718, opened 2023-04-20
updated section about if java 19 is accidentally installed

jennyowen:
I moved the section to AFTER installation instead of the java pre-requisites because it's something that will only happen *after* neo4j is installed.


renetapopova (approved):
Thanks, @jennyowen!

neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-718.surge.sh

PR #721, opened 2023-04-20
Add neo4j-admin validate-config

renetapopova:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-721.surge.sh

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
```diff
@@ -0,0 +1,62 @@
+:description: How to validate configurations using Neo4j Admin.
+[[neo4j-admin-validate-config]]
+= Validate configurations
+
+The `neo4j-admin server validate-config` command validates the Neo4j and Log4j configurations.
+It is located in the `bin` directory and is used to validate the configuration files before starting Neo4j.
+The command is available in Neo4j v5.5 and later.
```
lidiazuin, line 7: Are we mentioning the minor versions now? Because when 5 was released there was that discussion that we should only refer to Neo4j 5, no matter what minor version

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
```diff
@@ -0,0 +1,62 @@
...
+The `neo4j-admin server validate-config` command has the following syntax:
+
+[source,role=noheader]
+----
+neo4j-admin server validate-config [-h] [--expand-commands] [--verbose]
+----
+
+== Options
+
+The `neo4j-admin server validate-config` command has the following options:
```
lidiazuin, line 20: Should the command be repeated every time for each section?

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+.`neo4j-admin server validate-config` options
```
lidiazuin, line 22: I asked the previous question because this is getting very repetitive and long, it could give the impression that this page is longer and more time-consuming than it actually is

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
```diff
@@ -0,0 +1,62 @@
...
+[options="header", cols="5m,6a,2m"]
+|===
+| Option
+| Description
+| Default
```
lidiazuin, line 27: None of the 3 options have a value for the Default column?

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+|--expand-commands
+|Allow command expansion in config value evaluation.
```
lidiazuin, line 30:
```suggestion
|Allows command expansion in config value evaluation.
```

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
```diff
@@ -0,0 +1,62 @@
...
+|
+
+|-h, --help
+|Show this help message and exit.
```
lidiazuin, line 34:
```suggestion
|Shows this help message and exit.
```

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
```diff
@@ -0,0 +1,62 @@
...
+|
+|--verbose
+|Enable verbose output.
```
lidiazuin, line 37:
```suggestion
|Enables verbose output.
```

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
```diff
@@ -0,0 +1,62 @@
...
+|
+|===
+
+== Example
+
+.Invoke `neo4j-admin validate-conifg`
+[source, shell]
+----
+bin/neo4j-admin validate-conifg
```
lidiazuin, line 46:
```suggestion
bin/neo4j-admin validate-config
```

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
```diff
@@ -0,0 +1,62 @@
...
+|Show this help message and exit.
+|
+|--verbose
+|Enable verbose output.
+|
+|===
+
+== Example
+
+.Invoke `neo4j-admin validate-conifg`
```
lidiazuin, line 43: I would remove this title as it is repetitive. If you want to keep it, just consider the typo: validate-conifg -> validate-config

lidiazuin (changes requested):
Thanks Reneta! I found a few things that you might take in consideration 😊

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
renetapopova: Yes, we mention the minor version when a new feature is introduced or updated. Otherwise, the users will wonder why they don't have the feature even though they are on version 5.

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
renetapopova: Yes, because nobody is reading documentation as a book. Each section should be self-contained. So if a user is looking for the options of this command, they will not read the previous section.

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
renetapopova: About the captions, I don't have a strong opinion. We can decide to remove them or remove the command from them. Whatever we choose should be replicated on all commands' pages.

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
renetapopova: Good catch! Thanks!

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
renetapopova: Those descriptions are as they appear in the code as we want to test them. So, I tried only to change them if there was a mistake.

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
renetapopova: Thanks!

modules/ROOT/pages/tools/neo4j-admin/validate-config.adoc
lidiazuin: No, never mind then. I was just wondering :)

lidiazuin (approved):

PR #747, opened 2023-04-26
Removing sentence as requested

lidiazuin:
It alone does not communicate much meaning, so removing this for now as more context is provided.


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-747.surge.sh

PR #750, opened 2023-04-27
 Improve Configuration settings page

renetapopova:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-750.surge.sh

davidoliverSP2 (commented):
Just an observation, and perhaps something to look at outside of this PR...

The code block header is immediately repeated in the table name (+ the table numbering is pretty useless here due to the number of tables). I wonder if we can suppress the table numbering + table name on this page?

Screenshot 2023-04-27 at 20 46 50

An additional UI improvement would also be to right-justify the Enterprise Edition label.

modules/ROOT/pages/configuration/configuration-settings.adoc
```diff
@@ -0,0 +1,4477 @@
...
+* <<Logging settings>>
+* <<Memory settings>>
+* <<Metrics settings>>
+* <<Neo4j Browser and client settings>>
+* <<Kubernetes settings>>
+* <<Security settings>>
+* <<Server directories>>
+* <<Server settings>>
+* <<Transaction settings>>
+* <<Transaction logs settings>>
```
davidoliverSP2, line 38:
```suggestion
* <<Transaction log settings>>
```

modules/ROOT/pages/configuration/configuration-settings.adoc
```diff
@@ -0,0 +1,4477 @@
...
+
+== Checkpoint settings
+
+Checkpointing is the process of flushing all pending page updates from the page cache to the store files.
+This is done periodically and is used to recover the database in case of a crash.
+The checkpoint settings control the frequency of checkpoints, and the amount of data that is written to disk in each checkpoint.
+//For more information, see <<Checkpointing>>.
+See also, <<Transaction logs settings>>.
```
davidoliverSP2, line 46:
```suggestion
See also, <<Transaction log settings>>.
```

modules/ROOT/pages/configuration/configuration-settings.adoc
```diff
@@ -0,0 +1,4477 @@
+[[configuration-settings]]
+= Configuration settings
+:description: This page provides a complete reference to the Neo4j configuration settings.
+:page-styles: hide-table-captions
+
+This page provides a complete reference to the Neo4j configuration settings, which can be set in xref::/configuration/file-locations.adoc#file-locations[_neo4j.conf_].
+Refer to xref::/configuration/neo4j-conf.adoc#neo4j-conf[The neo4j.conf file] for details on how to use configuration settings.
+
+Some of the settings are labeled Dynamic, which means that they can be changed at runtime, without restarting the service.
```
davidoliverSP2, line 9:
```suggestion
Some of the settings are labeled *Dynamic*, which means that they can be changed at runtime, without restarting the service.
```

modules/ROOT/pages/configuration/configuration-settings.adoc
```diff
@@ -0,0 +1,4477 @@
...
+[[config_db.checkpoint]]
+
+[[config_db.checkpoint]]
+=== `db.checkpoint`
+
+.db.checkpoint
+[frame="topbot", stripes=odd, grid="cols", cols="<1s,<4"]
+|===
+|Description
+a|Configures the general policy for when checkpoints should occur. The default policy is the 'periodic' checkpoint policy, as specified by the '<<config_db.checkpoint.interval.tx,db.checkpoint.interval.tx>>' and '<<config_db.checkpoint.interval.time,db.checkpoint.interval.time>>' settings. The Neo4j Enterprise Edition provides two alternative policies: The first is the 'continuous' checkpoint policy, which will ignore those settings and run the checkpoint  process all the time. The second is the 'volumetric' checkpoint policy, which makes a best-effort at checkpointing often enough so that the database doesn't get too far behind on deleting old transaction logs in accordance with the '<<config_db.tx_log.rotation.retention_policy,db.tx_log.rotation.retention_policy>>' setting.
```
davidoliverSP2, line 57: Some of the descriptions have single quotes instead of backticks, so that setting names are not marked up correctly - I've amended them in this example, but there are a few others (for example, `dbms.cluster.minimum_initial_system_primaries_count`) I know these are matched to the content in the product - are you planning an editorial review of the descriptions, as I think paragraph breaks would be useful here too?

```suggestion
a|Configures the general policy for when checkpoints should occur. The default policy is the 'periodic' checkpoint policy, as specified by the `<<config_db.checkpoint.interval.tx,db.checkpoint.interval.tx>>` and `<<config_db.checkpoint.interval.time,db.checkpoint.interval.time>>` settings. The Neo4j Enterprise Edition provides two alternative policies: The first is the 'continuous' checkpoint policy, which will ignore those settings and run the checkpoint  process all the time. The second is the 'volumetric' checkpoint policy, which makes a best-effort at checkpointing often enough so that the database doesn't get too far behind on deleting old transaction logs in accordance with the `<<config_db.tx_log.rotation.retention_policy,db.tx_log.rotation.retention_policy>>` setting.
```

davidoliverSP2 (approved):
Thanks @renetapopova - I've a few editorial suggestions, including fixing a broken link to the transaction log settings.

This is an important step towards integrating the config settings into the configuration section. I really like the reorganisation, and how the labelling now works. I'm looking forward to us not needing an appendix at all for this documentation!

modules/ROOT/pages/configuration/configuration-settings.adoc
renetapopova: Yes, reviewing all descriptions is on my list.

PR #779, opened 2023-05-12
Polish server side routing information.

stefano-ottolenghi:
From Petr,

> The fact that server-side-routing uses Java driver is an internal detail. One cluster member forwards the query to another cluster member. The fact that it uses Java driver to do that is an implementation detail.

that's the reason for the change in the first sentence.

Sister PR: https://github.com/neo4j/docs-operations/pull/780

PR #780, opened 2023-05-12
Polish server side routing information and settings.

stefano-ottolenghi:
From Petr,

> The fact that server-side-routing uses Java driver is an internal detail. One cluster member forwards the query to another cluster member. The fact that it uses Java driver to do that is an implementation detail.

that's the reason for the change in the first sentence.

Sister PR: https://github.com/neo4j/docs-operations/pull/779


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-780.surge.sh

davidoliverSP2 (approved):
Looks good, thanks @stefano-ottolenghi 

👍

PR #778, opened 2023-05-12
Update ssl-framework.adoc

ali-ince:
I think we should mention to include the certificate chain in the `public.crt` file for clients to be able to verify them properly.

PR #786, opened 2023-05-16
Update ssl-framework.adoc (#778)

renetapopova:
I think we should mention to include the certificate chain in the `public.crt` file for clients to be able to verify them properly.

---------

Cherry-picked from #778


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-786.surge.sh

PR #800, opened 2023-05-22
Briefly describe each config group

renetapopova:


parnmatt:
[For review https://github.com/neo4j/docs-operations/pull/800](https://trello.com/c/NWcOEIQG/8773-for-review-https-githubcom-neo4j-docs-operations-pull-800)

renetapopova:
> The changes look great, but isn't at least part of the configuration settings page (like setting descriptions) auto-generated from the setting definitions in the Neo4j source (e.g. [here](https://github.com/neo4j/neo4j/blob/5ad4387ed521f169a737f9836402dbac8759a9fc/community/configuration/src/main/java/org/neo4j/configuration/GraphDatabaseSettings.java#L78))? Should some of these changes be made there instead?

No, they are not auto-generated anymore, but at the moment, they are pretty much the same as in the source. The next thing to do on this project is to review all descriptions and the other config fields, and together with @stefano-ottolenghi, we are also working on tests to be able to catch when there is an update to a setting. So a lot more to come in that direction.

PR #805, opened 2023-05-23
Add compatibility guarantee for composite databases

smietana:
We are expected to support compatibility with minor versions only, but will not actively prevent major version combinations.

This applies to all 5.x and later series.

### DIFF
--- a/modules/ROOT/pages/composite-databases/index.adoc
+++ b/modules/ROOT/pages/composite-databases/index.adoc
@@ -22,6 +22,9 @@ Composite databases are managed using Cypher administrative commands.
 Therefore, you can create them as any other database in Neo4j standalone and cluster deployments without needing to deploy a dedicated proxy server (as with Fabric in Neo4j 4).
 For a detailed example of how to create a Composite database and add database aliases to it, see xref:composite-databases/administration.adoc[Managing Composite databases].
 
+Composite databases cannot guarantee compatibility between constituents in major releases. Therefore, it's recommended that all constituents belong to the same major release family.
+If a new feature is introduced, its availability will be limited to the intersection of features available for all constituents in the composite.
+
 The following table summarizes the similarities and differences between Composite databases in Neo4j 5 and Fabric in Neo4j 4:
 
 .Composite database Neo4j 5 vs. Fabric Neo4j 4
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-805.surge.sh

modules/ROOT/pages/composite-databases/index.adoc
```diff
@@ -22,6 +22,9 @@ Composite databases are managed using Cypher administrative commands.
 Therefore, you can create them as any other database in Neo4j standalone and cluster deployments without needing to deploy a dedicated proxy server (as with Fabric in Neo4j 4).
 For a detailed example of how to create a Composite database and add database aliases to it, see xref:composite-databases/administration.adoc[Managing Composite databases].
 
+Composite databases cannot guarantee compatibility between constituents in major releases. Therefore, it's recommended that all constituents belong to the same major release family.
```
LinneaAndersson, line 25: Maybe it would be good to define "constituents" before we have this text? In the introduction we seem to only mention "aliases". Wdyt?

smietana: Good idea, updated

modules/ROOT/pages/composite-databases/index.adoc
```diff
@@ -15,13 +15,16 @@ It allows you to efficiently manage your data, computing resources, storage, and
...
-They use _database aliases_, which can be local or remote.
+_Database aliases_ are used to specify the constituents of the composite, they can be local or remote.
 For more information, see link:{neo4j-docs-base-uri}/cypher-manual/{page-version}/aliases/[Database alias management^].
 
 Composite databases are managed using Cypher administrative commands.
 Therefore, you can create them as any other database in Neo4j standalone and cluster deployments without needing to deploy a dedicated proxy server (as with Fabric in Neo4j 4).
 For a detailed example of how to create a Composite database and add database aliases to it, see xref:composite-databases/administration.adoc[Managing Composite databases].
 
+Composite databases cannot guarantee compatibility between constituents in major releases. Therefore, it's recommended that all constituents belong to the same major release family.
+If a new feature is introduced, its availability will be limited to the intersection of features available for all constituents in the composite.
```
renetapopova, line 26:
```suggestion
If a new feature is introduced, its availability will be limited to the intersection of features available for all constituents.
```

renetapopova (changes requested):
Thanks, @smietana! I tried to improve the wording and keep it coherent with the use of the concept fabric. Please, take a look and see if it makes it better.

renetapopova (approved):
Approved 👍

PR #809, opened 2023-05-24
Update minimum value for minimum system primaries

nick-giles-neo:
Setting this to 1 didn't work, so the constraint has been changed in https://github.com/neo-technology/neo4j/pull/20735.


nick-giles-neo:
The enforcement of this change will be in 5.9, but the docs update can apply to all of 5.9, since it would not have worked to set it to 1 before anyway.

PR #810, opened 2023-05-24
Update minimum value for minimum system primaries (#809)

AlexicaWright:
Setting this to 1 didn't work, so the constraint has been changed in https://github.com/neo-technology/neo4j/pull/20735.


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-810.surge.sh

PR #820, opened 2023-06-01
Add compatibility guarantee for composite databases (#805)

renetapopova:
We are expected to support compatibility with minor versions only, but will not actively prevent major version combinations.

This applies to all 5.x and later series.

---------
Cherry-picked from #805


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-820.surge.sh

PR #836, opened 2023-06-12
Admonition message about the applicability of the option

lidiazuin:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-836.surge.sh

renetapopova (approved):

renetapopova:
LGTM!  Thanks, @lidiazuin! Just wondering, my understanding is that we always work on the `dev` branch and add a label into which version it should go. So, if that's the current version, then we add the `5.x` label + `cherry-pick-this-to-5.x`.

lidiazuin:
> LGTM! Thanks, @lidiazuin! Just wondering, my understanding is that we always work on the `dev` branch and add a label into which version it should go. So, if that's the current version, then we add the `5.x` label + `cherry-pick-this-to-5.x`. Perhaps, we should clarify this again, as I might be wrong.

I'm not sure about that either! I remember hearing something about raising PRs on dev and then cherry-pick to 5.x and backwards, but then I can't know for sure if that's correct or not. Let's take that to Slack! 😁 And thanks for your review

PR #838, opened 2023-06-14
Improve the checkpointing configurations descriptions and remove +++

renetapopova:
@stefano-ottolenghi, could you take a look and see if these changes would work for the testing?

### DIFF
--- a/modules/ROOT/pages/configuration/configuration-settings.adoc
+++ b/modules/ROOT/pages/configuration/configuration-settings.adoc
@@ -64,7 +64,7 @@ A checkpoint is a point in the transaction logs, which recovery would start from
 Longer checkpoint intervals typically mean that recovery will take longer to complete in case of a crash.
 On the other hand, a longer checkpoint interval can also reduce the I/O load that the database places on the system, as each checkpoint implies a flushing and forcing of all store files.
 |Valid values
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-838.surge.sh

stefano-ottolenghi:
I have only looked at the first handful of settings, but I have noticed there have been significant changes to the descriptions and to valid values, in a way that makes their content different from the product, and it's not a matter of simple whitespace or formatting. Unless you also plan to submit those changes into Neo4j's source code, so that docs-in-the-product and these docs become equal again, we won't be able to test those fields. We can still check that all settings are present in the docs, but we cannot validate details.

renetapopova:
> I have only looked at the first handful of settings, but I have noticed there have been significant changes to the descriptions and to valid values, in a way that makes their content different from the product, and it's not a matter of simple whitespace or formatting. Unless you also plan to submit those changes into Neo4j's source code, so that docs-in-the-product and these docs become equal again, we won't be able to test those fields. We can still check that all settings are present in the docs, but we cannot validate details. You may already be aware of this and happy with that, but I just wanted to ensure we are on the same page :)

Of course. I know that we should also change the descriptions in the code base. Otherwise, we can't test them. Maybe we should have a quick chat about this PR.

renetapopova:
Replaced by #1062

PR #834, opened 2023-06-09
Add new constraint types to intro table

Hunterness:
- Added relationship key constraints as they were missing (node key existed)
- Added property uniqueness constraints cause there were none of them
- Added the new property type constraints (introduced in https://github.com/neo4j/docs-cypher/pull/597)

PR #839, opened 2023-06-14
Add new constraint types to intro table (#834)

AlexicaWright:
- Added relationship key constraints as they were missing (node key existed)
- Added property uniqueness constraints cause there were none of them
- Added the new property type constraints (introduced in https://github.com/neo4j/docs-cypher/pull/597)

---------


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-839.surge.sh

PR #851, opened 2023-06-16
Mark incremental import as enterprise only

emmaholmbergohlsson:
Has always been enterprise only, but it is not obvious from just looking at the command docs.

### DIFF
--- a/modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
+++ b/modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
@@ -399,6 +399,7 @@ If importing to a database that has not explicitly been created prior to the imp
 ====
 
 
+[role=enterprise-edition]
 [[import-tool-incremental]]
 == Incremental import
 


emmaholmbergohlsson:
[#13183 Unable to perform incremental import: No batch importers found](https://trello.com/c/iVcmxWD0/8717-13183-unable-to-perform-incremental-import-no-batch-importers-found)

neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-851.surge.sh

emmaholmbergohlsson:
Thanks @renetapopova
The incremental import was added in 5.0 I think, so it should be cherry-picked to whatever branches it is needed on for that. Wasn't sure about the correct labels
Will you handle merging to all applicable places as usual or should I do something more?

renetapopova:
> Thanks @renetapopova The incremental import was added in 5.0 I think, so it should be cherry-picked to whatever branches it is needed on for that. Wasn't sure about the correct labels Will you handle merging to all applicable places as usual or should I do something more?

We just need to cherry-pick it into 5.x branch. Could you be able to do that?

emmaholmbergohlsson:
@renetapopova of course :)
#852

PR #852, opened 2023-06-16
Mark incremental import as enterprise only

emmaholmbergohlsson:
Has always been enterprise only, but it is not obvious from just looking at the command docs.


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-852.surge.sh

renetapopova:
Thanks, @emmaholmbergohlsson!

PR #859, opened 2023-06-20
Improvements to the neo4j-admin copy documentation

renetapopova:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-859.surge.sh

modules/ROOT/pages/backup-restore/copy-database.adoc
```diff
@@ -235,19 +241,24 @@ Therefore, it will not be visible if you do `SHOW DATABASES` at this point.
...
-[source, cypher]
+[source, cypher, role=noplay]
 ----
-CREATE DATABASE copy
+CREATE DATABASE database-copy
 ----
++
+[TIP]
+====
+If you want to change the database name, you can rename both directories _data/databases/database-copy_ and _data/transactions/database-copy_ to _data/databases/new-database-name_ and _data/transactions/new-database-name_, and run the `CREATE DATABASE new-database-name` command.
```
emmaholmbergohlsson, line 251: This feels like a weird tip to me. If they wanted a different name they should have specified it when using the copy command when they started?
I don't think we should recommend them copying directories when it is not necessary, it is not something we want them to do

renetapopova: Thanks, @emmaholmbergohlsson. I agree with you. It sounded odd to me as well. That's why I decided to check it with you. I'll add your comment to the card, as this was the suggested workaround.

renetapopova: Hey @emmaholmbergohlsson, how can we preserve the database name in 4.4? Maybe this tip for renaming folders applies to the 4.4 version?

renetapopova: Ok, thanks, @emmaholmbergohlsson! I'll raise a PR against the 4.4. You can approve or disapprove of the change there, but at least we'll have proper traction.

PR #866, opened 2023-06-22
Add a page about the reconciler pattern in a clustered DBMS

nick-giles-neo:
This was in the 'Clustering internals' page in 4.4, but needed updating for 5.x, and making more useful.

### DIFF
--- a/modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
+++ b/modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
@@ -3,7 +3,6 @@
 = Reconciler
 :description: This section describes how changes to the DBMS are processed by each server.
 
-[[cluster-reconciler-introduction]]
 == Introduction
 
 In Neo4j, administrative operations such as database creation do not happen synchronously.
@@ -14,12 +13,10 @@ When the reconciler on `server-1` becomes aware of that, it starts `bar`.
 
 Executing an operation and getting back a successful response means that the desire is safely committed to the `system` database, and will be processed by each member of the cluster at some point, assuming the server is healthy enough.
 
-[[cluster-reconciler-async]]
 == Asynchronicity
 Servers can become aware the same operation at different times, since the `system` database is just another Raft group, where followers and replicas can lag behind the leader in some situations.
 Each server should eventually see the new state though, and take action on it.
 
-[[cluster-reconciler-wait]]
 == Waiting
 
 If you want to be sure that each server has processed an administrative operation before running the next action, you can use the link:{neo4j-docs-base-uri}/cypher-manual/current/administration/databases/#administration-wait-nowait[`WAIT`] keyword.
@@ -27,7 +24,7 @@ If you want to be sure that each server has processed an administrative operatio
 
 [NOTE]
 ====
-A statements with `WAIT` returning does not mean that your operation is necessarily successful everywhere.
+When a statement with `WAIT` returns, it does not mean that the operation was necessarily successful everywhere.
 `WAIT` just ensures the operation has been processed.
 For example, a request to start a database is considered _processed_ if the reconciler tried to start it, but the database failed and went into the dirty state.
 ====
@@ -38,7 +35,6 @@ For example, a request to start a database is considered _processed_ if the reco
 CREATE DATABASE foo WAIT
 ----
 
-[[cluster-reconciler-errors]]
 == Errors
 
 An operation might only succeed on some servers.
@@ -50,7 +46,6 @@ Other failures only require the underlying problem to be resolved.
 When the issue is fixed, the reconciler will make the intended change, because it is still trying to achieve the desired state.
 For example, if a new server fails during a `DEALLOCATE DATABASES FROM SERVER` (meaning a database cannot be safely moved), fixing the new server should be enough to resolve the problem, as the new server will start its copy of the database as soon as the target is ready, allowing the deallocating server to shut down its copy.
 
-[[cluster-reconciler-reason]]
 == Why use the reconciler pattern
 
 This approach is robust to more failures.
--- a/modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
+++ b/modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
@@ -1,0 +1,51 @@
+[role=enterprise-edition]
+[[cluster-reconciler]]
+= Reconciler
+:description: This section describes how changes to the DBMS are processed by each server.
+
+[[cluster-reconciler-introduction]]
+== Introduction
+
+Administrative operations in Neo4j, such as creating a database, do not happen synchronously.
+Instead we have a model where the changes are written to the `system` database, and record the desired state of the DBMS.
+Each server has an internal component called the reconciler, which will then make any changes relevant to that server once it observes that desire, for example starting a local copy of a database because it has been allocated to them.
+Servers can observe the same operation at different times, since the `system` database is just another Raft group, where followers and replicas can lag behind the leader in some situations.
+Executing an operation and getting back a successful response means that the desire is safely committed to the `system` database, and will be processed by each member of the cluster at some point, assuming the server is healthy enough.
+
+[[cluster-reconciler-wait]]
+== Waiting
+
+If you want to be sure that each server has processed an administrative operation before you run the next action, you can use the `WAIT` keyword.
+This makes the statement not return until all servers have seen the transaction, and their reconciler has finished processing it.
+
+[NOTE]
+====
+This does not mean that your operation is necessarily successful everywhere.
+`WAIT` just ensures the operation has been processed.
+A request to start a database has still been processed if the reconciler tried to start it and the database failed and went into the dirty state.
+====
+
+=== Example using `WAIT`
+[source, cypher, role="noplay"]
+----
+CREATE DATABASE foo WAIT
+----
+
+[[cluster-reconciler-errors]]
+== Errors
+
+An operation might only succeed on some servers.
+For instance some servers might be offline when you create a database, or a disk error might cause a server to fail to start a database.
+In this situation the operation has not 'failed' as a whole, since you may even have a running database, just with less fault tolerance than you desired.
+
+Some failures can just be re-attempted, for example `START DATABASE` can be run if not all members started successfully, and you think the reason has been resolved.
+Other failures will need just the underlying problem to be resolved and the reconciler will make the change, because it is still trying to achieve the desired state.
+For example if a new server fails during a `DEALLOCATE DATABASES FORM SERVER`, meaning a database cannot be safely moved, fixing the new server should be enough to resolve the problem, as the new server will start its copy of the database, allowing the deallocating server to shut down its copy.
+
+[[cluster-reconciler-reason]]
+== Why use the reconciler pattern
+
+This approach is robust to more failures.
+Each server is responsible for getting its local state to match its copy of the global desire.
+This means the loss of any one server cannot prevent all progress on that operation.
+If there was for instance one coordinator making changes to all other servers, a failure of that coordinator would result in progress stopping until the failure could be detected, and a replacement started.
--- a/modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
+++ b/modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
@@ -6,18 +6,16 @@
 == Introduction
 
 In Neo4j, administrative operations such as database creation do not happen synchronously.
-Instead, the changes are first written to the `system` database, recording the desired state of the DBMS.
-Each server has a _reconciler_, an internal component which observes the desired state and makes changes to the local server to match that state.
+Instead, the changes are first written to the `system` database, recording the requested state of the DBMS.
+Each server has a _reconciler_, an internal component which observes the requested state and makes changes to the local server to match that state.
 For example, the `system` database may record that the database `bar` is expected to be running on `server-1`.
 When the reconciler on `server-1` becomes aware of that, it starts `bar`.
+This means the loss of any one server cannot prevent all progress on an operation.
 
-Executing an operation and getting back a successful response means that the desire is safely committed to the `system` database, and will be processed by each member of the cluster at some point, assuming the server is healthy enough.
+Executing an operation and getting back a successful response means that the request is safely committed to the `system` database, and will be processed by each member of the cluster at some point, assuming the server is healthy enough.
 
-== Asynchronicity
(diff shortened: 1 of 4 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-866.surge.sh

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
+[role=enterprise-edition]
+[[cluster-reconciler]]
+= Reconciler
+:description: This section describes how changes to the DBMS are processed by each server.
+
+[[cluster-reconciler-introduction]]
+== Introduction
+
+Administrative operations in Neo4j, such as creating a database, do not happen synchronously.
```
stefano-ottolenghi, line 9:
```suggestion
In Neo4j, administrative operations such as database creation do not happen synchronously.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+Instead we have a model where the changes are written to the `system` database, and record the desired state of the DBMS.
```
stefano-ottolenghi, line 10:
```suggestion
Instead, the changes are first written to the `system` database, recording the desired state of the DBMS.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+Each server has an internal component called the reconciler, which observes the desired state and makes changes to the local server to match that state.
```
stefano-ottolenghi, line 11:
```suggestion
Each server has a _reconciler_, an internal component which observes the desired state and makes changes to the local server to match that state.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+For example, the `system` database may record that it wants the database `bar` running on `server-1`.
```
stefano-ottolenghi, line 12:
```suggestion
For example, the `system` database may record that the database `bar` is expected to be running on `server-1`.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+When the reconciler on `server-1` sees that, it will start `bar`.
```
stefano-ottolenghi, line 13:
```suggestion
When the reconciler on `server-1` becomes aware of that, it starts `bar`.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+Executing an operation and getting back a successful response means that the desire is safely committed to the `system` database, and will be processed by each member of the cluster at some point, assuming the server is healthy enough.
```
stefano-ottolenghi, line 15:
```suggestion
Executing an operation and getting back a successful response means that the desired state is safely committed to the `system` database, and that it will be processed by each member of the cluster at some point (assuming the server is healthy enough).
```
`desire` as a verb or noun is _very_ strong https://dictionary.cambridge.org/dictionary/english/desire
The adjective is okay, but I recommend reviewing the whole text and checking if another adjective is possible. For example, how about _intended state_?

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
+[role=enterprise-edition]
+[[cluster-reconciler]]
+= Reconciler
+:description: This section describes how changes to the DBMS are processed by each server.
+
+[[cluster-reconciler-introduction]]
```
stefano-ottolenghi, line 6: Are all the section labels needed? Do you want specific IDs for each of them? Applies to later sections as well.

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+Administrative operations in Neo4j, such as creating a database, do not happen synchronously.
+Instead we have a model where the changes are written to the `system` database, and record the desired state of the DBMS.
+Each server has an internal component called the reconciler, which observes the desired state and makes changes to the local server to match that state.
+For example, the `system` database may record that it wants the database `bar` running on `server-1`.
+When the reconciler on `server-1` sees that, it will start `bar`.
+
+Executing an operation and getting back a successful response means that the desire is safely committed to the `system` database, and will be processed by each member of the cluster at some point, assuming the server is healthy enough.
+
+[[cluster-reconciler-async]]
+== Asynchronicity
```
stefano-ottolenghi, line 18: I recommend taking away both this and the next section and just incorporating these two paragraphs into the main text. Separate sections for 2 sentences each is quite short.

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+Servers can observe the same operation at different times, since the `system` database is just another Raft group, where followers and replicas can lag behind the leader in some situations.
```
stefano-ottolenghi, line 19:
```suggestion
Servers can become aware the same operation at different times, since the `system` database is just another Raft group, where followers and replicas can lag behind the leader in some situations.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+Each server should eventually see the new state though, and take action on it.
```
stefano-ottolenghi, line 20: This remark leaves readers with doubt. _Are there circumstances where the new state is not propagated? Do I need to be careful and account for it? Are there known issues on this?_
Can we make it affirmative, as in 

> At all rates/Sooner or later, the new state gets eventually propagated to all servers.

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+[[cluster-reconciler-wait]]
+== Waiting
+
+If you want to be sure that each server has processed an administrative operation before you run the next action, you can use the `WAIT` keyword.
```
stefano-ottolenghi, line 25:
```suggestion
If you want to be sure that each server has processed an administrative operation before running the next action, you can use the link:{neo4j-docs-base-uri}/cypher-manual/current/administration/databases/#administration-wait-nowait[`WAIT`] keyword.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+`WAIT` makes the statement not return until all servers have seen the transaction, and their reconciler has finished processing it.
+
+[NOTE]
+====
+A statements with `WAIT` returning does not mean that your operation is necessarily successful everywhere.
```
stefano-ottolenghi, line 30:
```suggestion
When a statements with `WAIT` returns, it does not mean that the operation was necessarily successful everywhere.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+`WAIT` just ensures the operation has been processed.
+A request to start a database has still been processed if the reconciler tried to start it and the database failed and went into the dirty state.
```
stefano-ottolenghi, line 32:
```suggestion
For example, a request to start a database is considered _processed_ if the reconciler tried to start it, but the database failed and went into the dirty state.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+====
+
+=== Example using `WAIT`
+[source, cypher, role="noplay"]
```
stefano-ottolenghi, line 36:
```suggestion
[source, cypher]
```

The play button has been taken away some months back so `noplay` has no effect anymore.

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+`WAIT` makes the statement not return until all servers have seen the transaction, and their reconciler has finished processing it.
+
+[NOTE]
+====
+A statements with `WAIT` returning does not mean that your operation is necessarily successful everywhere.
+`WAIT` just ensures the operation has been processed.
+A request to start a database has still been processed if the reconciler tried to start it and the database failed and went into the dirty state.
+====
+
+=== Example using `WAIT`
```
stefano-ottolenghi, line 35:
```suggestion
.Create database with `WAIT`
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+[source, cypher, role="noplay"]
+----
+CREATE DATABASE foo WAIT
+----
+
+[[cluster-reconciler-errors]]
+== Errors
+
+An operation might only succeed on some servers.
+For instance some servers might be offline when you create a database, or a disk error might cause a server to fail to start a database.
```
stefano-ottolenghi, line 45:
```suggestion
For instance, some servers might be offline when a database is created, or a disk error might cause a server to fail to start a database.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+In this situation the operation has not 'failed' as a whole, since you may even have a running database, just with less fault tolerance than you desired.
```
stefano-ottolenghi, line 46:
```suggestion
In this situation, the operation has not _failed_ as a whole, since the database may even be running (although with less fault tolerance than you intended).
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+Some failures can just be re-attempted, for example `START DATABASE` can be run if not all members started successfully, and you think the reason has been resolved.
```
stefano-ottolenghi, line 48:
```suggestion
Some failed operations can just be re-tried: for example, `START DATABASE` can be re-run if not all members started successfully, and you think the cause of that has been resolved.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+Other failures will need just the underlying problem to be resolved and the reconciler will make the change, because it is still trying to achieve the desired state.
```
stefano-ottolenghi, line 49:
```suggestion
Other failures only require the underlying problem to be resolved. 
When the issue is fixed, the reconciler will make the intended change, because it is still trying to achieve the desired state.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+For example if a new server fails during a `DEALLOCATE DATABASES FORM SERVER`, meaning a database cannot be safely moved, fixing the new server should be enough to resolve the problem, as the new server will start its copy of the database, allowing the deallocating server to shut down its copy.
```
stefano-ottolenghi, line 50:
```suggestion
For example, if a new server fails during a `DEALLOCATE DATABASES FROM SERVER` (meaning a database cannot be safely moved), fixing the new server should be enough to resolve the problem, as the new server will start its copy of the database as soon as the target is ready, allowing the deallocating server to shut down its copy.
```

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+[[cluster-reconciler-reason]]
+== Why use the reconciler pattern
```
stefano-ottolenghi, line 53: Do we need to justify to customers _why_ we use this design? Do we really need this section? Also, I believe any tech person realizes that a distributed architecture is more robust than a centralized one :)
But if we really do, I'd recommend pruning the section a bit and merging it with the introduction (i.e. taking away the section header).

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
nick-giles-neo: We use `requestedHosting` in `SHOW SERVERS`, _intended_ is fine by me but _requested_ is an existing term, so probably preferable?

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
nick-giles-neo: The doubt is sort-of deliberate. There _are_ circumstances where a server will not see the results, though they are examples of bad problems. The server being permanently offline, long network partitions etc.

I could add a different qualifier, like:
> Eventually, the new state gets propagated to all healthy servers.

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
nick-giles-neo: Ah, I thought I was supposed to label all sections. I don't think deep linking to a section on a page this small is necessarily important, happy to remove them.

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
stefano-ottolenghi: > Eventually, the new state gets propagated to all healthy servers.

This sounds great.

modules/ROOT/pages/clustering/clustering-advanced/reconciler.adoc
stefano-ottolenghi: Yeah, _requested_ is also fine!

stefano-ottolenghi:
Now looking at the final preview, I would

- move the `WAIT` example before the note admonition
- reconsider whether we need the `Introduction` header (which ended up containing the largest part of the content). If we don't, we can also add `:page-toclevels: 0` in the document header to remove the ToC which would otherwise have only one entry

stefano-ottolenghi (approved):

PR #891, opened 2023-07-05
Document how to move from single to cluster

jackwaudby:
This PR adds documentation on how a user can move from a deployment with a single system primary server to a cluster deployment.

### DIFF
--- a/modules/ROOT/pages/clustering/setup/single-to-cluster.adoc
+++ b/modules/ROOT/pages/clustering/setup/single-to-cluster.adoc
@@ -1,20 +1,18 @@
-:description: This section describes how to move from a single Neo4j server to Neo4j cluster.
+:description: This section describes how to move from a single `system` primary deployment to a cluster.
 [role=enterprise-edition]
 [[single-to-cluster]]
-= Move from single to cluster
+= Move from a single `system` primary to a cluster deployment
 
-This is an example of how to move from a single Neo4j server to a Neo4j cluster. 
+This is an example of how to move from a deployment with a single `system` primary to a cluster with multiple `system` primaries.
 
-== Before you begin
-. Ensure that the single server is stopped. 
-
-== Migration steps 
-. On the single server add `dbms.cluster.discovery.endpoints` to `neo4j.conf`.
-. Setup the others server and configure them as per xref:clustering/setup/deploy.adoc[Configure a cluster with three servers].
-. On the single server dump the `system` database using `neo4j-admin database dump`.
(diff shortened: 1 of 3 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-891.surge.sh

modules/ROOT/pages/clustering/setup/single-to-cluster.adoc
```diff
@@ -0,0 +1,18 @@
+:description: This section describes how to move from a single `system` primary deployment to a cluster.
+[role=enterprise-edition]
+[[single-to-cluster]]
+= Move from a single `system` primary to a cluster deployment
```
AlexicaWright, line 4: It feels counterintuitive to talk about a _primary_ in a non-cluster deployment.

modules/ROOT/pages/clustering/setup/single-to-cluster.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+This is an example of how to move from a deployment with a single `system` primary to a cluster with multiple `system` primaries.
+
+== Guide
```
AlexicaWright, line 8: This doesn't follow the structure elsewhere in the clustering chapter. I will take a stab at rewriting it tomorrow.

AlexicaWright (commented):
Some comments and also I will push a suggested rewrite directly tomorrow (Friday 7/7)

modules/ROOT/pages/clustering/setup/single-to-cluster.adoc
jackwaudby: So my use of the term "single system primary" and thus avoidance of the terms "standalone" or just "single" came from some internal discussion (and tbh the lack of clarity I have around the official names of the various deployments we support.) 

The reason for using "single system primary deployment" was that we have the "analytics cluster" deployment, where there is a single system primary but several secondaries. Calling that primary a "single" or "standalone" somewhat ignores the existence of the secondaries. 

Having read your rewrite I'm all for using "standalone" but I do think we need to acknowledge somewhere that standalone can mean just 1 server on its own or 1 primary server with N secondaries. I guess the docs would have to say you would need to discard your secondaries from your old "analytics cluster".

AlexicaWright: Thank you for clarifying that! I wasn't sure, but now you've confirmed that the _single system primary_ indeed includes both standalone and analytical cluster setups, I will add a note about this.

PR #896, opened 2023-07-07
added debian 12 to the list of supported operating systems

jennyowen:
Debian 12 was released recently, and so we should support it.

PR #898, opened 2023-07-07
added debian 12 to the list of supported operating systems (#896)

renetapopova:
Debian 12 was released recently, and so we should support it.
Cherry-picked from #896


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-898.surge.sh

PR #900, opened 2023-07-11
Add new column `composite` to `SHOW ALIASES`

Hunterness:
Companion PR to https://github.com/neo4j/docs-cypher/pull/649.

Column added in 5.11 and I assume docs is currently pointing to 5.10 so I'll leave this as a draft PR until 5.10 is released and docs branch switched.

### DIFF
--- a/modules/ROOT/pages/composite-databases/administration.adoc
+++ b/modules/ROOT/pages/composite-databases/administration.adoc
@@ -83,12 +83,12 @@ The link:{neo4j-docs-base-uri}/cypher-manual/{page-version}/aliases#alias-manage
 SHOW ALIASES FOR DATABASE
 ----
 ----
-+----------------------------------------------------------------------------------------+
-| name                | database     | location | url                        | user      |
-+----------------------------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/composite-databases/administration.adoc
```diff
@@ -83,12 +83,12 @@ The link:{neo4j-docs-base-uri}/cypher-manual/{page-version}/aliases#alias-manage
 SHOW ALIASES FOR DATABASE
```
Hunterness, line 83: Not related to my update but the sentence on line 76 feels more like it belongs with this example and should be under the same header and not before the header?

l-heemann (approved):

modules/ROOT/pages/composite-databases/administration.adoc
JPryce-Aklundh: Thanks, @Hunterness - we will soon reorganise this page (as part of the migration from Cypher Manual), and will take this into consideration when we get to it  (unless @renetapopova wants to take action before).

JPryce-Aklundh (approved):

PR #904, opened 2023-07-14
fix retrieving serverId command for seed cluster section

ikwattro:
The section for retrieving the serverId here https://neo4j.com/docs/operations-manual/current/clustering/databases/#cluster-designated-seeder is not correct.

The docs use `SHOW SERVERS` which will return the name of the server, by default it has the same name as the serverId, but in case servers were renamed, using 
```
RENAME SERVER `25a7efc7-d063-44b8-bdee-f23357f89f01` TO `primary1` 
```
then it will not match anymore and trying to use `primary1` as serverId while following the documentation will lead to the following exception when seeding a cluster :
```
Could not create database with specified existingDataSeedInstance 'optimus-001'. Expected server uuid string.
```

This PR ensures that the correct serverId is returned when following the documentation.

### DIFF
--- a/modules/ROOT/pages/clustering/databases.adoc
+++ b/modules/ROOT/pages/clustering/databases.adoc
@@ -112,18 +112,18 @@ Use any database to connect.
 +
 [source, cypher, role=noplay]
 ----
-SHOW SERVERS;
+SHOW SERVERS YIELD serverId, name, address, state, health, hosting;
 ----
 +
 [queryresult]
 ----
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-904.surge.sh

AlexicaWright (approved):
Yes, this makes sense. Thank you @ikwattro !

PR #905, opened 2023-07-14
Use HTML codes for "<EOF >" to prevent interpretation as HTML element

kylecombes:
## The Problem

Currently, the documentation generator is not doing any escaping of the `<` and the `>` in the code snipped, so browsers are interpreting `<EOF >` an HTML element tag:

Screenshot 2023-07-06 at 12 02 49 PM

Thus, "&lt;EOF &gt;" is not being rendered:

Screenshot 2023-07-06 at 12 01 06 PM

This is a problem because copying and pasting the code snippet as it is rendered (without the `<EOF >`) does not work because the shell doesn't recognize the multi-line input as one command.

## The Fix

Escaping the carets using HTML `&gt;` and `&lt;` codes leads to proper rendering of the code snippet:

Screenshot 2023-07-14 at 4 52 56 PM

### DIFF
--- a/modules/ROOT/pages/installation/linux/rpm.adoc
+++ b/modules/ROOT/pages/installation/linux/rpm.adoc
@@ -55,7 +55,7 @@ To use the repository for generally available versions of Neo4j, run the followi
 [source, shell, subs="attributes"]
 ----
 rpm --import https://debian.neo4j.com/neotechnology.gpg.key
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-905.surge.sh

PR #907, opened 2023-07-17
Removing obsolete encryption arguments

lidiazuin:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-907.surge.sh

modules/ROOT/pages/security/ssl-framework.adoc
```diff
@@ -169,26 +169,16 @@ openssl pkcs8 -topk8 -v2 aes-256-cbc -v2prf hmacWithSHA512 -in pkcs1or8.key -out
...
-* `-v1 PBE-SHA1-RC2-128`
-* `-v1 PBE-SHA1-RC4-40`
-* `-v1 PBE-SHA1-RC4-128`
-* `-v1 PBE-SHA1-2DES`
 * `-v2 aes-128-cbc -v2prf hmacWithSHA1`
 * `-v2 aes-128-cbc -v2prf hmacWithSHA224`
 * `-v2 aes-128-cbc -v2prf hmacWithSHA256`
 * `-v2 aes-128-cbc -v2prf hmacWithSHA384`
 * `-v2 aes-128-cbc -v2prf hmacWithSHA512`
-* `-v2 aes-256-cbc -v2prf hmacWithSHA1`
```
irene221b, line 184: I think hmacWithSHA1 is still ok for some usecases. Maybe we can keep it with a caveat?

irene221b: Like, use carefully, the secrecy of the key is paramount, be sure that your use case doesn't rely on collision resistance etc. 

Or maybe we should just follow NIST advice and not mention it, you are right. :)

lidiazuin: I was thinking it would be better to just remove them at all and keep only the ones that are the safest. But in case there's any occasion where hmacWithSHA1 is mandatory or the most logical option, we can bring it back and leave a note? Also, is this valid for all versions of Neo4j?

irene221b: Yes, it is for all the versions. You are right, let's remove it and see if anyone asks for it!

PR #911, opened 2023-07-18
Improve the logging page

renetapopova:


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-911.surge.sh

modules/ROOT/pages/monitoring/logging.adoc
```diff
@@ -1,215 +1,334 @@
...
 The monitor interval tells Log4j to periodically check the XML file for changes and reload the file if a change is detected. +
 The package namespace gives access to the Neo4j configuration lookup with `${config:<setting>}`.
 
-| 3, 13
-| Defines two `<RollingFile>` appenders.
-  One writes to _debug.log_ and one to _http.log_.
-  The `name` must be unique and is used later when referencing the appender.
+<2> Appenders are used to define where the log messages are written.
+The `name` of the appender must be unique as it is used by the logger when referencing the appender.
+Appenders write to _debug.log_, _http.log_, _query.log_, and _security.log_.
```
MishaDemianenko, line 265: Default appenders

modules/ROOT/pages/monitoring/logging.adoc
```diff
@@ -1,215 +1,334 @@
...
 
-| 4, 14
-| Specifies a `filePattern` to be used when the file is rolled.
-  The pattern renames the files to _debug.log.01_ and _http.log.01_ when they reach the defined trigger.
+<3> By default, Neo4j uses the `<RollingRandomAccessFile>` appender as it shows a performance increase between 20-200% as per the Log4j documentation.
```
MishaDemianenko, line 267: Not sure ```as it shows a performance increase between 20-200% as per the Log4j documentation.``` is relevant here

modules/ROOT/pages/monitoring/logging.adoc
```diff
@@ -1,215 +1,334 @@
...
+However, all writes are buffered, which means that if the server crashes, the last log messages might be lost.
+If that is not acceptable for you, use the `<RollingFile>` appender instead.
+See <<general-logging-appenders>> for more information.
 
-| 5
-| Defines the layout to use for the `DebugLog` appender, in this case with the `GMT+2` timezone.
+<4> `filePattern` specifies a file pattern to be used when the file is rolled.
```
MishaDemianenko, line 272: For items 4 and 5 I would add a link to log4j docs since they are referencing log4j specific attributes in the body of patterns

modules/ROOT/pages/monitoring/logging.adoc
```diff
@@ -1,215 +1,334 @@
...
 
-.An example of a console appender
-[source,xml]
-----
-<Console name="console" target="SYSTEM_OUT"> <!-- or SYSTEM_ERR -->
-  <PatternLayout pattern="%m%n"/>
-</Console>
-----
+==== `<RollingRandomAccessFile>` appender
+The `<RollingRandomAccessFile>` is the default appender in Neo4j.
```
MishaDemianenko, line 314: a lof of info in this section is a duplication from the previous section with numbers 1-11. Is that intentional?

modules/ROOT/pages/monitoring/logging.adoc
```diff
@@ -375,6 +506,56 @@ There are also a couple of built-in templates available from the classpath, for
...
+[source,xml]
+----
+<RollingRandomAccessFile name="QueryLog" fileName="${config:server.directories.logs}/query.log"
+                                 filePattern="$${config:server.directories.logs}/query.log.%02i">
+            <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSSZ}{GMT+0} %-5p %m%n"/>
+            <Policies>
+                <SizeBasedTriggeringPolicy size="20 MB"/>
+            </Policies>
+            <!-- Filter out logs for queries executed by "neo4j" user -->
+            <Filters>
```
MishaDemianenko, line 523: if someone will do that on the log level - that's not really nice no one will ever see those queries over there and any support case maintainer will gonna be puzzled, so not sure we need to advertise this feature

MishaDemianenko (commented):
Have a couple of comments here and there, not sure if we change the grand picture of things in regards to this page but will let that for the original requestor to answer

modules/ROOT/pages/monitoring/logging.adoc
renetapopova: Yeah, maybe not the best. I just couldn't come up with a better justification why the default is <RollingRandomAccessFile>.

modules/ROOT/pages/monitoring/logging.adoc
renetapopova: Yes, I did it because nobody reads the whole page. So in this way, the Appenders section is self-contained and compact if they happen to read just this section.

modules/ROOT/pages/monitoring/logging.adoc
renetapopova: Could you be able to help me with some examples of using Filters and Plugins?

modules/ROOT/pages/monitoring/logging.adoc
renetapopova: I added a link to the section where there is also a link to the log4j docs.

modules/ROOT/pages/monitoring/logging.adoc
MishaDemianenko: Do we really need to provide justification for that? it's not like for example we describe why setting default has some particular value.

renetapopova: This is from the card:
"Given the warning for buffers will be lost on crash for that [RandomAccessFile](https://neo4j.com/docs/operations-manual/current/monitoring/logging/#_rollingrandomaccessfile_appender) can we have a normal RollingFile instead?"

MishaDemianenko: So before switching to log4j we were using a homemade framework that was using async log events - with the possibility of losing messages as well, and that was never a problem for anyone, for example.

modules/ROOT/pages/monitoring/logging.adoc
MishaDemianenko: So for example `Neo4jDebugLogLayout` that is mentioned in the docs is in fact `plugin`. 

So filter and plugins are more advanced things that by default neo4j users should not update/touch. Was there a specific request why we need to include them? My thinking is that if you are already at the stage where you want to write your own plugin or use a custom filter - you would be able to find that in log4j docs/examples.

MishaDemianenko (approved):

PR #917, opened 2023-07-20
Adding information about port 7688

lidiazuin:
The SSL Framework page (https://neo4j.com/docs/operations-manual/5/security/ssl-framework/ ) does not mention port 7688 (Server Side Routing).

This port would be encrypted if you use the cluster policy (or Bolt).
We should add that to the documentation to eliminate any possible confusion.

### DIFF
--- a/modules/ROOT/pages/security/ssl-framework.adoc
+++ b/modules/ROOT/pages/security/ssl-framework.adoc
@@ -7,7 +7,7 @@ The SSL framework provides support for securing the following Neo4j communicatio
 
 * `bolt` (port - `7687`)
 * `https` (port - `7473`)
-* `cluster` (ports - `5000`, `6000`, and `7000`)
+* `cluster` (ports - `5000`, `6000`, `7000`, and `7688`)
 * `backups` (port - `6362`)
 
 This page describes how to set up SSL within your environment, how to view, validate, and test the certificates.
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-917.surge.sh

modules/ROOT/pages/security/ssl-framework.adoc
```diff
@@ -271,7 +271,7 @@ Each policy needs to be explicitly enabled by setting:
 === Configure SSL over Bolt
 
 Bolt protocol is based on the link:https://neo4j.com/docs/bolt/current/packstream/[PackStream serialization] and supports the Cypher type system, protocol versioning, authentication, and TLS via certificates.
-For Neo4j clusters, Bolt provides smart client routing with load balancing and failover.
+For Neo4j clusters, Bolt provides smart client routing with load balancing and failover through port 7688.
```
nick-giles-neo, line 274: This addition feels slightly misleading, though technically accurate in some senses. 'Smart client routing' and 'load balancing' do not use server side routing, and 'failover' doesn't either in most cases (if a secondary or follower fails a query, send it to a different one mentioned in the routing table). Server-side routing is an additional Bolt port that is only for use within the cluster.

Maybe something along the lines of 'When server side routing is enabled, there will be an additional Bolt port open on 7688 for use only within the cluster, using all the same settings as the external Bolt port'?

lidiazuin: Thanks Nick. I have added your suggestion

nick-giles-neo (approved):
Looks good to me.

smietana (commented):
👌 looks good

PR #919, opened 2023-07-21
Add support for PVC labels

renetapopova:
To be cherry-picked when Helm Charts 5.10 are released.

### DIFF
--- a/modules/ROOT/pages/kubernetes/persistent-volumes.adoc
+++ b/modules/ROOT/pages/kubernetes/persistent-volumes.adoc
@@ -63,7 +63,8 @@ By default, the Neo4j Helm chart uses a single PV, named `data`, to support volu
 
 The volume used for each volume mount can be changed by modifying the `volumes.<volume name>` object in the Helm chart values.
 
-The Neo4j Helm chart `volumes` object supports different modes:
+The Neo4j Helm chart `volumes` object supports different modes, such as `dynamic`, `share`, `defaultStorageClass`, `volume`, `selector`, and `volumeClaimTemplate`.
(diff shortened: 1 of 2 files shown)


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-919.surge.sh

modules/ROOT/pages/kubernetes/persistent-volumes.adoc
```diff
@@ -158,6 +162,8 @@ When this method is used, the `persistentVolumeClaim` object must already exist.
 ----
 volumes:
   backups:
+  labels:
+        backup: true
     mode: volume
```
harshitsinghvi22, line 167: here the mode is type volume so the label will not work

harshitsinghvi22: we can remove the labels from here

renetapopova: aaah, good catch!

modules/ROOT/pages/kubernetes/persistent-volumes.adoc
```diff
@@ -82,6 +83,8 @@ neo4j:
   name: standalone-with-storage-class
 volumes:
   data:
+    labels:
+        data: true
```
harshitsinghvi22, line 87: can we also quote the values "true" ...i believe without quotes the example might just fail...so better to quote the value in documentation

renetapopova: But that's a boolean. Are you sure?

harshitsinghvi22: yeah since its a label...labels are considered strings

renetapopova: aah, ok.

harshitsinghvi22 (approved):

PR #930, opened 2023-07-27
Info about skipping consistency check for performance improvement on …

lidiazuin:
…backup

### DIFF
--- a/modules/ROOT/pages/backup-restore/planning.adoc
+++ b/modules/ROOT/pages/backup-restore/planning.adoc
@@ -215,3 +215,14 @@ If you have a cluster, you should back up these files for each cluster member.
 For any backup, it is important that you store your data separately from the production system, where there are no common dependencies, and preferably off-site.
 If you are running Neo4j in the cloud, you may use a different availability zone or even a separate cloud provider.
 Since backups are kept for a long time, the longevity of archival storage should be considered as part of backup planning.
+
+[[performance-improvement]]
+== Performance improvement
+
+Backups in Neo4j automatically run a consistency check against the backed-up store. 
+The backup itself does not take overly long, but the consistency check can take much more time to complete, depending on the store.
+
+In case you want to skip the check, with the option to run it manually at some other time, run the neo4j-backup tool with the `-verify false` option:
+----
+./bin/neo4j-backup -from single://127.0.0.1 -to /tmp/foo -verify false
+----
--- a/modules/ROOT/pages/backup-restore/planning.adoc
+++ b/modules/ROOT/pages/backup-restore/planning.adoc
@@ -222,7 +222,4 @@ Since backups are kept for a long time, the longevity of archival storage should
 Backups in Neo4j automatically run a consistency check against the backed-up store. 
 The backup itself does not take overly long, but the consistency check can take much more time to complete, depending on the store.
 
-In case you want to skip the check, with the option to run it manually at some other time, run the neo4j-backup tool with the `-verify false` option:
-----
-./bin/neo4j-backup -from single://127.0.0.1 -to /tmp/foo -verify false
-----
+In case you want to skip the check, with the option to run it manually at some other time, refer to the xref:tools/neo4j-admin/consistency-checker.adoc[Consistency checker] page.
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/backup-restore/planning.adoc
```diff
@@ -214,4 +214,15 @@ If you have a cluster, you should back up these files for each cluster member.
...
+
+[[performance-improvement]]
+== Performance improvement
+
+Backups in Neo4j automatically run a consistency check against the backed-up store. 
+The backup itself does not take overly long, but the consistency check can take much more time to complete, depending on the store.
+
+In case you want to skip the check, with the option to run it manually at some other time, run the neo4j-backup tool with the `-verify false` option:
+----
+./bin/neo4j-backup -from single://127.0.0.1 -to /tmp/foo -verify false
```
renetapopova, line 227: I can't find this command and the verify option in the command help.

renetapopova (changes requested):
Hey @lidiazuin, thanks for picking this up. Who told you that command and option? I can't find it anywhere.

modules/ROOT/pages/backup-restore/planning.adoc
lidiazuin: I got this from an [old developer guide](https://neo4j.com/developer/kb/how-can-i-skip-consistency-check-during-backup/#:~:text=In%20some%20cases%2C%20you%20want,Copied!), but I thought of asking someone from engineering to take a look to see whether this is still valid or not

renetapopova: Ah, yes. It says "Applicable versions 3.5 4.0 4.1 4.2 4.3 4.4" but also has "Retired:" in the title. I'll check the code. You can ask the Clustering team since they are the owners of the backup and restore if we have anything similar in 5.x.

neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-930.surge.sh

modules/ROOT/pages/backup-restore/planning.adoc
renetapopova: So, in 4.4, when I run `./neo4j-admin backup --help`, I see that there is an option `--check-consistency` (`true` by default). So if you want to skip the consistency check, you can add `--check-consistency false` to the backup command and it'll skip it. This is described here: https://neo4j.com/docs/operations-manual/4.4/backup-restore/online-backup/#online-backup-resources

lidiazuin: Right, thanks for checking it. In the 5.x version of this same page, there's nothing about the consistency check though. I have asked the Clustering team to review this PR, but I guess I can open another one for 4.x linking to this section that you suggested instead?

renetapopova: There is nothing about it on both 4.x and 5.x planning pages because the planning is about backing up in general (all possible ways). The consistency check is performed only on the backups that are created with the online back command. Therefore, the section about consistency checks is on the Back up an online database page. I don't think we need anything more than what we currently have. Perhaps, we can add a link to the consistency checker page - https://neo4j.com/docs/operations-manual/5/tools/neo4j-admin/consistency-checker/, in case they have skipped it during the backup and want to run it later.

tonbut: @lidiazuin @renetapopova the consistency checking functionality has been completely removed from the backup command in 5.x. I believe this is because it is not possible to perform a consistency check on a diff backup. So the only way not to perform the consistency check now is offline using tool as documented here https://neo4j.com/docs/operations-manual/5/tools/neo4j-admin/consistency-checker/

tonbut: So the correct process now is to either perform a consistency check periodically on your full backups or use the aggregate command to combine diff backups together with a full to create a complete full backup (all this can be done offline). The run the consistency check on the complete aggregated backup.

lidiazuin: Thanks @tonbut . I believe with your information now this PR can be actually closed and we can work only in the 4.x one, in case it is still valid.

PR #941, opened 2023-07-31
Clarify that database status endpoints only work for local databases

nick-giles-neo:
Databases hosted on other servers and not the local one do not have status endpoints on the local server, so you get `404`.


nick-giles-neo:
This has been true for all of 5.x, so can be merged whenever.

AlexicaWright (approved):
LGTM!

PR #970, opened 2023-08-14
Add vector index procedures

parnmatt:


parnmatt:
Related changes in Cypher manual: https://github.com/neo4j/docs-cypher/pull/690

recrwplay:
The workflow uses `npm install --omit-dev` to install only the packages in `package.json` that are required to generate HTML. Moving the package from `devDependencies` into `dependencies` will work.

However, the extension is not supported by PDF builds so the examples won't be correctly rendered. To exclude the examples from PDF or to include an alternative (ie svg) you can use a condition:

```
ifndef::backend-pdf[]
// content to exclude from PDF goes here
endif::[]

ifdef::backend-pdf[]
// content to include only in PDF goes here
endif::[]
```

parnmatt:
So what I'm hearing is that it'll make everyone's life easier if I pre-render an SVG 😅

parnmatt:
Ok, I'll work on producing them at some point soon.

Hopefully you can still render the html locally using `npm install` and `npm start` etc.

…So I'm not stalling anything on the rest of the content review 😁

renetapopova:
I'll move the package locally and try again.

renetapopova (approved):
I managed to build it locally and it looks good! Thanks, @parnmatt!

neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-970.surge.sh

modules/ROOT/pages/reference/procedures.adoc
```diff
@@ -907,6 +939,58 @@ m|READ
...
+    := \dfrac1{1 + \left\lVert\vec v - \vec u\right\rVert^2}
+++++
++
+
+* `"cosine"` -- Cosine similarity, where:
++
+[latexmath]
+++++
+\begin{align}
+    \mathrm{cosign}(\vec v,\vec u)
```
nilsceberg, line 966: Typo in `cosine`, and I would probably clarify / make explicit the unit vector notation :)

parnmatt: ah bugger, I thought I got all of those.
I'll fix that in the SVG render I need to make.

Well the unit vector notation is clarified on the next line of maths :wink: 
However, I don't think it needs too much more clarification as those that are likely double-checking the mathematical calculation of a score from a vector-based thing … likely know the basics of vectors.

nilsceberg: > likely know the basics of vectors

Well, like most mathematical notation it's less standard / ubiquitous than one would've liked, so I wouldn't count on it :P

parnmatt: Thoughts on that for context? It's the same in both.

parnmatt:
> Thanks for being ok with a workaround for now @parnmatt. The question of supporting mathjax has come up before, but not recently. It would be better if we could support it fully so it might be time to look at updating our PDF generation.

No problem, it is what it is :smile: 
As we're hosting, KaTeX can be faster, but really only makes a difference if there are a lot of maths on a page.
Though I do not believe it has an SVG output. Apparently MathJax SVG output works well in browsers, and _some_ viewers, but not all … thus it may not embed well into a PDF :man_shrugging: 

So I chose to be on the safer side and go from LaTeX to PDF to SVG to ensure the structures should work.

modules/ROOT/pages/reference/procedures.adoc
nilsceberg: Perfect :+1:

parnmatt:
ah, I was going to change a line of text after everything was done with cypher.
I'll make another small PR tomorrow then 😄

PR #977, opened 2023-08-15
Document index metrics

nilsceberg:
Admittedly it's not very pretty to enumerate each metric like this because they are identical except for the index type, but that's how they're defined in the software. Thoughts?


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-977.surge.sh

renetapopova (approved):
Looks good! Thanks, @nilsceberg!

PR #1002, opened 2023-08-24
add a clustering glossary for 4.4

AlexicaWright:


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1002.surge.sh

modules/ROOT/pages/clustering/glossary.adoc
```diff
@@ -0,0 +1,123 @@
...
+The bookmark is passed along and updated by all subsequent queries in the session, regardless of which server executes what query.
+A bookmark can only be updated monotonically increasing.
+If a server is behind the state in the bookmark, it waits until it has caught up, or time out the query.
+Thus, clients executing queries within a session are guaranteed to read their own writes, and only see successively later states of the database.
+This is sometimes also referred to as <<glossary-session-consistency,session consistency>>.
+
+[[glossary-cluster]]
+| Causal cluster
+| A collection of servers running Neo4j that are configured to communicate with each other.
+The servers can be either _Primary_ or _Secondary_, where the Primary servers allow read and write operations and the Secondary servers allow only read operations.
```
gfx54b, line 41: Generally the question is if we would not want to use the terms "core" and "read-replica". I think in case of a 4.4 they are completely fine, probably even better as the rest of the docs, would mention those, and not primary and secondary.

Same for below for _Primary server_ and _Secondary server_

modules/ROOT/pages/clustering/glossary.adoc
```diff
@@ -0,0 +1,123 @@
...
+The cluster guarantees that a client application can read at least its own writes.
```
gfx54b, line 42: The cluster guarantees that a client application can read at least its own writes. -- if the client uses bookmarks.

modules/ROOT/pages/clustering/glossary.adoc
```diff
@@ -0,0 +1,123 @@
...
+The number of failures `f` that can be tolerated is dependent on the number of primaries `n` for the database and follows the formula `f = (n-1)/2`.
+In the event that more than `f` primaries fail, the database can no longer process write transactions and becomes read-only.
+
+[[glossary-follower]]
+| Follower
+| A primary copy of a database acting as a follower, receives and acknowledges synchronous writes from the leader.
+
+[[glossary-leader]]
+| Leader
+| A single Primary server is designated as the leader.
```
gfx54b, line 75: Leadership is per database. Different cores can be leaders for different databases. The 5.x glossary entry is fine here.

modules/ROOT/pages/clustering/glossary.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+[[glossary-secondary]]
+| Secondary server
+| An asynchronously replicated instance that provides read scaling within the cluster.
+Secondary servers are made up of Read Replica instances.
+
+
+[[glossary-seed]]
+| Seed
+| A file used to create a database on a single instance or on a member of a cluster.
```
gfx54b, line 98: I see this is also how it is expressed in the 5 glossary, but it feels weird to me to say the seed is *a* file. It is a bunch of files, it is is a whole store, even in different directories. I think we should leave this alone now in this PR, but return to it later, and fix it for both gloassaries.

modules/ROOT/pages/clustering/glossary.adoc
AlexicaWright: But in 5.x, we say _A single primary copy of a database_, this concept doesn't really apply in 4.4, does it? Also, the last sentence in the description does express that leadership is per database.

modules/ROOT/pages/clustering/glossary.adoc
AlexicaWright: Agreed, this can be rephrased, but I'll leave it for now.

modules/ROOT/pages/clustering/glossary.adoc
AlexicaWright: Thanks, I've amended.

gfx54b (approved):

PR #1063, opened 2023-09-28
Update procedures with new set procedure and beta status

parnmatt:
* drop beta tag vector index create and query procedures
* keep beta tag on set property procedure
* deprecate old set procedure in favor of new procedure with different name and signature


parnmatt:
Unsure how best to "tag" _was_ beta, due to these docs being for all 5.x
I have gone with "Beta until 5.13", rather than keeping "Beta" or removing it … it's they are beta for 5.11 and 5.12, but not for 5.13+

neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1063.surge.sh

renetapopova (approved):

PR #1064, opened 2023-09-28
clarify that on Neo4j startup all pagecache warmup activity results in pagefaults

renetapopova:


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1064.surge.sh

modules/ROOT/partials/neo4j-metrics/available-metrics.adoc
```diff
@@ -104,7 +104,7 @@
 |&lt;prefix&gt;.page_cache.pins|The total number of page pins executed by the page cache. (counter)
 |&lt;prefix&gt;.page_cache.evictions|The total number of page evictions executed by the page cache. (counter)
 |&lt;prefix&gt;.page_cache.evictions.cooperative|The total number of cooperative page evictions executed by the page cache due to low available pages. (counter)
-|&lt;prefix&gt;.page_cache.page_faults|The total number of page faults happened in the page cache. If this continues to rise over time, it could be an indication that more page cache is needed. (counter)
+|&lt;prefix&gt;.page_cache.page_faults|The total number of page faults happened in the page cache. If this continues to rise over time, it could be an indication that more page cache is needed. Be aware that on Neo4j startup, all page cache warmup activity results in page faults and so it is not unexpected to see a significant page fault count immediately after startup. (counter)
```
emmaholmbergohlsson, line 107: Page cache warmup is an enterprise feature, and also controllable by a setting (but on by default) so not applicable for everyone. The number of page faults also depends on both store size and page cache size so I'm not sure if significant is the correct word all the time, but probably okay.
How about a small change to:
`Be aware that page cache warmup results in page faults and so it is not unexpected to see a significant page fault count immediately after startup on Neo4j Enterprise.`

emmaholmbergohlsson: This text needs to be updated in the product code as well right? They should still be up to date even without the autogeneration?

renetapopova: Sorry for the delayed reply, @emmaholmbergohlsson. Given that I'll update the code base as well, how about rewording it completely to:
"The total number of page faults in the page cache. If this count keeps increasing over time, it may indicate that more page cache is required. However, note that when Neo4j Enterprise starts up, all page cache warmup activities result in page faults. Therefore, it is normal to observe a significant page fault count immediately after startup."

emmaholmbergohlsson: Sure, looks okay to me

emmaholmbergohlsson (approved):

PR #1083, opened 2023-10-10
Add parallel runtime information

JPryce-Aklundh:
The parallel runtime introduced a new config setting. For testing purposes, I have taken the description from the mono repo verbatim (https://github.com/neo-technology/neo4j/pull/22275/files#diff-22e5a6968aee5ed485e16da8650e784a16dbd787005216bb719d3dc44aa2b43a)

There are also a number of procedures that are not considered thread-safe to run on the parallel runtime. This PR adds a note for the relevant procedures. Note that the link to the Cypher Manual will not work yet as the we have yet to merge the parallel runtime PR in that repo.

### DIFF
--- a/modules/ROOT/pages/configuration/configuration-settings.adoc
+++ b/modules/ROOT/pages/configuration/configuration-settings.adoc
@@ -1595,6 +1595,30 @@ a|a double which is in the range `0.0` to `1.0`
 m|+++0.75+++
 |===
 
+
+[[config_server.cypher.parallel.worker_limit]]
+=== `server.cypher.parallel.worker_limit`
+
+label:enterprise-edition[Enterprise Edition]
+label:new[Introduced in 5.13]
+
+.server.cypher.parallel.worker_limit 
+[frame="topbot", stripes=odd, grid="cols", cols="<1s,<4"]
+|===
+|Description
+a| Number of threads to allocate to Cypher worker threads for the parallel runtime.
+If set to a positive number, that number of workers will be started.
+If set to `0`, one worker will be started for every logical processor available to the Java Virtual Machine.
+
+If set to a negative number we will subtract the value from the number of logical processors available; for example, say Neo4j is running on a server with 16 available processors, using `server.cypher.parallel.worker_limit = -1` would then mean that 15 threads are available for the parallel runtime.
+
+|Valid values
+a| Integer
+
+|Default value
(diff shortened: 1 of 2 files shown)


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1083.surge.sh

modules/ROOT/pages/configuration/configuration-settings.adoc
```diff
@@ -1595,6 +1595,30 @@ a|a double which is in the range `0.0` to `1.0`
...
+
+.server.cypher.parallel.worker_limit 
+[frame="topbot", stripes=odd, grid="cols", cols="<1s,<4"]
+|===
+|Description
+a| Number of threads to allocate to Cypher worker threads for the parallel runtime.
+If set to a positive number, that number of workers will be started.
+If set to `0`, one worker will be started for every logical processor available to the Java Virtual Machine.
+
+If set to a negative number we will subtract the value from the number of logical processors available; for example, say Neo4j is running on a server with 16 available processors, using `server.cypher.parallel.worker_limit = -1` would then mean that 15 threads are available for the parallel runtime.
```
renetapopova, line 1613:
```suggestion
If set to a negative number, that number will be subtracted from the total number of logical processors available on the server. For example, if the server has 16 available processors and you set `server.cypher.parallel.worker_limit` to `-1`, the parallel runtime will have 15 threads available.
```

renetapopova (commented):
I have one suggestion. Thanks, @JPryce-Aklundh.

modules/ROOT/pages/configuration/configuration-settings.adoc
JPryce-Aklundh: Thanks @renetapopova. I agree that it is slightly awkwardly phrased. However, I took it verbatim from the mono-repo, and changing it will mean there is a discrepancy between docs and the mono-repo. I propose that I change it in the mono-repo once 5.13 has been released and then update the docs accordingly (and that we therefore merge this as is). What do you think?

renetapopova: Sure.

renetapopova: I'll create a card so as not to forget and merge this one.

PR #1085, opened 2023-10-11
Update information on `db.transaction.timeout`

NataliaIvakina:
Following the slack discussions and findings, I suggest the following amendments.
Do we have then to cherry-pick this PR to the branch 5.x to make it visible now?

### DIFF
--- a/modules/ROOT/pages/database-internals/transaction-management.adoc
+++ b/modules/ROOT/pages/database-internals/transaction-management.adoc
@@ -57,7 +57,7 @@ Configuring transaction timeout does not affect transactions executed with custo
 Note that the timeout value can be overridden to a value smaller than that configured by `db.transaction.timeout`.
 
 label:new[Introduced in Neo4j 5.3]
-Starting from Neo4j 5.3 and onwards, you can set the transaction timeout to any value, even larger than configured by `db.transaction.timeout`.
+Starting from Neo4j 5.3, you can set the transaction timeout to any value, even larger than configured by `db.transaction.timeout`.
 
 
 == Manage transactions
--- a/modules/ROOT/pages/database-internals/transaction-management.adoc
+++ b/modules/ROOT/pages/database-internals/transaction-management.adoc
@@ -54,7 +54,10 @@ db.transaction.timeout=10s
 ====
 
 Configuring transaction timeout does not affect transactions executed with custom timeouts (e.g., via the Java API or Neo4j Drivers), as the custom timeout overrides the value set for `db.transaction.timeout`.
-Note that the timeout value can only be overridden to a value that is smaller than that configured by `db.transaction.timeout`.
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1085.surge.sh

modules/ROOT/pages/database-internals/transaction-management.adoc
```diff
@@ -54,7 +54,10 @@ db.transaction.timeout=10s
 ====
 
 Configuring transaction timeout does not affect transactions executed with custom timeouts (e.g., via the Java API or Neo4j Drivers), as the custom timeout overrides the value set for `db.transaction.timeout`.
-Note that the timeout value can only be overridden to a value that is smaller than that configured by `db.transaction.timeout`.
+Note that the timeout value can be overridden to a value smaller than that configured by `db.transaction.timeout`.
```
robsdedude, line 57: Without the only this sentence dons't make much sense to me. Reading up to here I'd think "Of course it can! If you don't tell me otherwise, I'll assume it can be overridden to anything".

It also leaves me wondering "okay, and what about values greater than `db.transaction.timeout`?"

modules/ROOT/pages/database-internals/transaction-management.adoc
```diff
@@ -54,7 +54,10 @@ db.transaction.timeout=10s
...
+
+label:new[Introduced in Neo4j 5.3]
+Starting from Neo4j 5.3 and onwards, you can set the transaction timeout to any value, even larger than configured by `db.transaction.timeout`.
```
robsdedude, line 60:
```suggestion
Starting from Neo4j 5.3, you can set the transaction timeout to any value, even larger than configured by `db.transaction.timeout`.
```

"Starting from" implies some sort of continuation, I'd think.

modules/ROOT/pages/database-internals/transaction-management.adoc
NataliaIvakina, line 57:
```suggestion
Note that the timeout value can only be overridden to a value smaller than that configured by `db.transaction.timeout`.
```

NataliaIvakina: ok, I see. Put the word 'only' back into the sentence. 
I have another question: should we clarify here that it's true only in versions 5.1 and 5.2?

robsdedude (approved):
:alarm_clock:

modules/ROOT/pages/database-internals/transaction-management.adoc
robsdedude: I think it's clear enough as well.

Furthermore, 4.4 behaves just like 5.1 on this regard and everything before 4.4 is EOL, so I'd argue it's not a big deal if the documentation here doesn't apply to e.g., 4.2.

PR #1115, opened 2023-10-17
added note about clearing yum cache on upgrade

jennyowen:
I will create a separate PR for the 4.4 documentation.

### DIFF
--- a/modules/ROOT/pages/installation/linux/rpm.adoc
+++ b/modules/ROOT/pages/installation/linux/rpm.adoc
@@ -64,6 +64,12 @@ gpgcheck=1
 EOF
 ----
 
+[NOTE]
+====
+If you are upgrading from Neo4j 4.4 or earlier, you may need to clear the package manager cache before neo4j packages become available:
+
+`yum clean dbcache`
+====
 
 === Install Neo4j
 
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1115.surge.sh

modules/ROOT/pages/installation/linux/rpm.adoc
```diff
@@ -64,6 +64,12 @@ gpgcheck=1
 EOF
 ----
 
+[NOTE]
+====
+If you are upgrading from Neo4j 4.4 or earlier, you may need to clear the package manager cache before neo4j packages become available:
```
renetapopova, line 69: If they a moving from 4.4 to 5.x., they need to migrate. Also, from an earlier version than 4.4, migration is not supported. I actually think that this note should be added to the U&M guide instead of here. 
```suggestion
If you are upgrading from Neo4j 4.4 or earlier, you may need to clear the package manager cache before Neo4j packages become available:
```

renetapopova (commented):
I added a comment. Thanks, @jennyowen

renetapopova (approved):

PR #1116, opened 2023-10-17
added note about clearing yum cache on upgrade

jennyowen:
also we should be including branch number in the script for adding our rpm repository (like we already do for 5x)

### DIFF
--- a/modules/ROOT/pages/installation/linux/rpm.adoc
+++ b/modules/ROOT/pages/installation/linux/rpm.adoc
@@ -35,7 +35,7 @@ EOF
 +
 [NOTE]
 ====
-If you are upgrading from an older version of Neo4j, you may need to clear the package manager cache before neo4j packages become available:
+If you are upgrading from an older version of Neo4j, you may need to clear the package manager cache before Neo4j packages become available:
 `yum clean dbcache`
 ====
 +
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1116.surge.sh

modules/ROOT/pages/installation/linux/rpm.adoc
```diff
@@ -21,18 +21,24 @@ However, Java supports pointing to an alternative `temp` directory by either set
...
-baseurl=https://yum.neo4j.com/stable
+baseurl=https://yum.neo4j.com/stable/{neo4j-version}
 enabled=1
 gpgcheck=1
 EOF
 ----
 +
+[NOTE]
+====
+If you are upgrading from an older version of Neo4j, you may need to clear the package manager cache before neo4j packages become available:
```
renetapopova, line 38:
```suggestion
If you are upgrading from an older version of Neo4j, you may need to clear the package manager cache before Neo4j packages become available:
```

renetapopova: I think they might not see this note if they are upgrading because all the upgrade information is in the Upgrade and Migration guide.

renetapopova (approved):
Thanks, @jennyowen. I added a small suggestion and a comment. Otherwise, looks good.

modules/ROOT/pages/installation/linux/rpm.adoc
renetapopova: I actually think that this note should be added to the U&M guide instead of here.

renetapopova (commented):
Please see my comment.

jennyowen:
@renetapopova perhaps the note should be both here *and* in the upgrade manual?

PR #1166, opened 2023-10-27
Docker - ref-settings.adoc: NEO4J_dbms_security_ldap_host typo error

MarlinMr:
I was getting an error when using the original variable name, but removing one `_` from each side of `ldap` seems to have fixed it. Can anyone else also confirm that?


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1166.surge.sh

jennyowen (approved):
thank you, yes that looks like a mistake. I think the one on line 500 is also incorrect.

PR #1183, opened 2023-11-14
Improve documentation for suspended users

mnd999:
Explain the effect of suspending a user that is currently using the system.

### DIFF
--- a/modules/ROOT/pages/authentication-authorization/manage-users.adoc
+++ b/modules/ROOT/pages/authentication-authorization/manage-users.adoc
@@ -6,6 +6,26 @@
 Users can be created and managed using a set of Cypher administration commands executed against the `system` database.
 When connected to the DBMS over `bolt`, administration commands are automatically routed to the `system` database.
 
+== User States
+Users in the system database can exist in two states, `ACTIVE` and `SUSPENDED`. `ACTIVE` is the default state for new users on
+the database and when a user is in the `ACTIVE` state they are able to log into the system and perform queries as usual according
+to their privileges.
+
+`SUSPENDED` users are not permitted to log into the system using the credentials in the system graph. This does not prevent them
+from logging in to the database using external authentication. To prevent this the user should be suspended there using the mechanisms
+provided by the identity system.
+
+In the event that the user is using the database at the point they are suspended:
+
+* A user authorizing against only the system graph will be stripped of all roles assigned in the system graph, including the `PUBLIC` role.
+This has the effect of removing all privileges.
+* If the user is assigned roles by an external authorization provider, these roles will remain assigned to that user, and they will also
+retain the `PUBLIC` role. As above, the user should also be suspended there using the mechanism provided by that system.
+
+[NOTE]
+====
+The `SUSPENDED` state is only available in Neo4j Enterprise Edition. label:enterprise-edition[]
+====
 
 [[access-control-user-syntax]]
 == User management command syntax
@@ -368,7 +388,7 @@ Will return `null` in community edition.
 | {check-mark}
 
 | suspended
-| If `true`, the user is currently suspended (cannot log in).
+| If `true`, the user is currently suspended.
 
 Will return `null` in community edition.
 | BOOLEAN
@@ -448,7 +468,7 @@ Will return `null` in community edition.
 | {check-mark}
 
 | suspended
-| If `true`, the user is currently suspended (cannot log in).
+| If `true`, the user is currently suspended.
 
 Will return `null` in community edition.
 | BOOLEAN
--- a/modules/ROOT/pages/authentication-authorization/manage-users.adoc
+++ b/modules/ROOT/pages/authentication-authorization/manage-users.adoc
@@ -6,26 +6,19 @@
 Users can be created and managed using a set of Cypher administration commands executed against the `system` database.
 When connected to the DBMS over `bolt`, administration commands are automatically routed to the `system` database.
 
-== User States
-Users in the system database can exist in two states, `ACTIVE` and `SUSPENDED`. `ACTIVE` is the default state for new users on
-the database and when a user is in the `ACTIVE` state they are able to log into the system and perform queries as usual according
-to their privileges.
+== User states
 
-`SUSPENDED` users are not permitted to log into the system using the credentials in the system graph. This does not prevent them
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1183.surge.sh

modules/ROOT/pages/authentication-authorization/manage-users.adoc
```diff
@@ -6,6 +6,26 @@
...
+`SUSPENDED` users are not permitted to log into the system using the credentials in the system graph. This does not prevent them
+from logging in to the database using external authentication. To prevent this the user should be suspended there using the mechanisms
+provided by the identity system.
+
+In the event that the user is using the database at the point they are suspended:
+
+* A user authorizing against only the system graph will be stripped of all roles assigned in the system graph, including the `PUBLIC` role.
+This has the effect of removing all privileges.
+* If the user is assigned roles by an external authorization provider, these roles will remain assigned to that user, and they will also
+retain the `PUBLIC` role. As above, the user should also be suspended there using the mechanism provided by that system.
```
Hunterness, line 23: should this list also include that we don't throw/log them out?

Hunterness: I'm not sure either 🤷

mnd999:
@renetapopova This came out of a bug where the removing roles behaviour was not removing the PUBLIC role, but we also found that the behaviour of SUSPEND was not well documented in the first place. 
Support are asking whether we should explicitly document the bug and the fixed version as well, which I have not done. Is this something we should consider also, or would we prefer to keep the docs clear of this level of detail.

modules/ROOT/pages/authentication-authorization/manage-users.adoc
```diff
@@ -6,6 +6,26 @@
 Users can be created and managed using a set of Cypher administration commands executed against the `system` database.
 When connected to the DBMS over `bolt`, administration commands are automatically routed to the `system` database.
 
+== User States
```
renetapopova, line 9: I started adding comments but then decided to add the whole suggestion in a comment as otherwise, it might become too patchy. So here it is:

```
== User states

There are two types of user states in the `system` database:

* `ACTIVE` state (the default for new users) -- users can log into Neo4j and perform queries according to their privileges.

* `SUSPENDED` state label:enterprise-edition[] -- users are not permitted to log into Neo4j using the credentials in the system graph. 
If they attempt to do so, they will lose all their assigned roles, including `PUBLIC`.
+
[NOTE]
====
Users with roles assigned by an external authorization provider can still access Neo4j by logging in with their external authentication credentials.
To prevent this, you must suspend their access by using the mechanisms provided by the identity system.
It is important to note that even if access is suspended, their roles will remain assigned, including the `PUBLIC` role.
====
```

renetapopova (changes requested):
Thanks, @mnd999. I've added an editorial suggestion.

modules/ROOT/pages/authentication-authorization/manage-users.adoc
Hunterness: > users are not permitted to log into Neo4j using the credentials in the system graph. 
If they attempt to do so, they will lose all their assigned roles, including `PUBLIC`.

Isn't really the case though. They can't log in at all if they are suspended, that part has nothing to do with the roles. If they already have an open connection we don't kick them out when they get suspended but it is in this open connection their roles will be removed (temporarily, if they get reactivated then the roles will still be there).

renetapopova: Ok, let me think again.

mnd999:
> Usually, we don't document bugs. We only mention them in the Release notes. But if you think this one is very important and should be mentioned, we can do that, of course.

I think it's fine not to, will pass that on to support.

modules/ROOT/pages/authentication-authorization/manage-users.adoc
renetapopova: Is it correct now?

````
== User states

There are two types of user states in the `system` database:

* `ACTIVE` state (the default for new users) -- users can log into Neo4j and perform queries according to their privileges.

* `SUSPENDED` state label:enterprise-edition[] -- users cannot log into Neo4j using the credentials in the system graph. 
If the users are suspended while using Neo4j, they lose all assigned roles, including `PUBLIC`, until reactivated.
+
[NOTE]
====
Users with external authentication and roles assigned by an external authorization provider must be suspended using the mechanisms of their identity provider. 
====

renetapopova: Ah, yes. I might have missed that.

mnd999: I started off structuring it the way you have it but then changed it because the suspended option is way more complicated than the active one and required more explanation. You certainly cleaned it up, but maybe too much. 
If a user is suspended whilst using Neo4j, they only lose the PUBLIC role if they have no other authorization. If they authorize against something else (e.g. LDAP) they will not lose the public role. This is to avoid breaking use cases where native users are being created and suspened to allow e.g. setting home database for that user when the log in with LDAP. The account would be suspended because we don't want to use the native credentials to sign in, but we do want to be able to store the home database somewhere.

renetapopova: Screenshot 2023-11-22 at 12 42 39

renetapopova: Screenshot 2023-11-22 at 12 44 16

renetapopova: @mnd999, how about now?

renetapopova: Any thoughts?

mnd999: Yeah, that works for me.

renetapopova: Perfect! Thanks, I'll update the PR.

renetapopova (approved):
Thanks, @mnd999!

PR #1215, opened 2023-11-28
Add new target for `SHOW SUPPORTED PRIVILEGES`

Hunterness:
added with `LOAD ON ALL DATA` but was missed when it was documented


renetapopova (approved):
Thanks, @Hunterness

l-heemann (approved):

PR #1218, opened 2023-11-29
Improve the description of normalize-types behavior

mauvo:


mauvo:
I didn't make a note about 5.14 inconsistent behaviour.

There's a link to Cypher types that needs proper AsciiDoc formatting.

neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1218.surge.sh

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
```diff
@@ -236,7 +236,7 @@ For an example, see <<import-tool-multiple-input-files-regex-example>>.
 |
 
 |--normalize-types[=true\|false]
-|Whether or not to normalize property types to Cypher types, e.g. `int` becomes `long` and `float` becomes `double`.
+|When true, non-array property values will be converted to their equivalent Cypher types. For example, all integer values will be converted to 64-bit long integers. This takes up more space on disk, but avoids Cypher converting the type during queries.
```
renetapopova, line 239:
```suggestion
|When `true`, non-array property values will be converted to their equivalent Cypher types. For example, all integer values will be converted to 64-bit long integers. This takes up more space on disk but avoids Cypher converting the type during queries.
```

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
```diff
@@ -605,8 +605,7 @@ For an example, see <<import-tool-multiple-input-files-regex-example>>.
 |
 
 |--normalize-types[=true\|false]
-|Whether or not to normalize property types to Cypher types, e.g. `int` becomes `long` and `float` becomes `double`.
-|true
+|When true, non-array property values will be converted to their equivalent Cypher types. For example, all integer values will be converted to 64-bit long integers. This takes up more space on disk, but avoids Cypher converting the type during queries.
```
renetapopova, line 608:
```suggestion
|When `true`, non-array property values will be converted to their equivalent Cypher types. For example, all integer values will be converted to 64-bit long integers. This takes up more space on disk but avoids Cypher converting the type during queries.
```

renetapopova: The descriptions must be updated in the code base as well. Otherwise, the test will fail.

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
```diff
@@ -921,10 +920,16 @@ You can have properties in both node data files and relationship data files.
 Data types::
 Use one of `int`, `long`, `float`, `double`, `boolean`, `byte`, `short`, `char`, `string`, `point`, `date`, `localtime`, `time`, `localdatetime`,
 `datetime`, and `duration` to designate the data type for properties.
+By default, types (except arrays) will be converted to Cypher types [https://neo4j.com/docs/cypher-manual/current/values-and-types/property-structural-constructed/#_property_types].
```
renetapopova, line 923:
```suggestion
By default, types (except arrays) are converted to Cypher types. 
See link:{neo4j-docs-base-uri}/cypher-manual/{page-version}/values-and-types/property-structural-constructed/#_property_types[Cypher Manual -> Values and Types].
```

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
```diff
@@ -921,10 +920,16 @@ You can have properties in both node data files and relationship data files.
...
+This behavior can be disabled using the option `--normalize-types=false`.
+Normalizing types can require more space on disk, but avoids Cypher converting the type during queries.
 If no data type is given, this defaults to `string`.
+
 To define an array type, append `[]` to the type.
 By default, array values are separated by `;`.
 A different delimiter can be specified with `--array-delimiter`.
+Arrays are affected by the `--normalize-types` option. If you want a byte array to be stored as a Cypher long array you must explicitly declare the property as `long[]`.
```
renetapopova, line 931:
```suggestion
Arrays are affected by the `--normalize-types` option.
If you want a byte array to be stored as a Cypher `long` array, you must explicitly declare the property as `long[]`.
```

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
brunobuss, line 239: @renetapopova do you know if these are pulled automatic from [our codebase](https://github.com/neo-technology/neo4j/blob/dbefc055901079cd05287fd8d99f1d7c9e67764c/public/community/import-tool/src/main/java/org/neo4j/importer/ImportCommand.java#L338)?

renetapopova (commented):
Some editorial suggestions. Thanks, @mauvo!

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
brunobuss, line 239: > This takes up more space on disk, but avoids Cypher converting the type during queries.

I think @tinwelint mentioned this might not be the case, no?

Did you managed to check this by testing different runs of import with normalization on/off and the difference on the disk-size used?

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
brunobuss, line 931:
```suggestion
Arrays are not affected by the `--normalize-types` flag. For example, if you want a byte array to be stored as a Cypher long array you must explicitly declare the property as `long[]`.
```

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
renetapopova: No, they are not. That's why @mauvo raised this PR here. However, we have a test that validates the neo4j-admin options. I added a comment a bit further above that we need to update the codebase as well when we agree on the descriptions.

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
mauvo: I should have written: "Arrays are not affected...". I accepted @brunobuss suggestion further down that corrects this but undoes the formatting changes @renetapopova made here.

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
tinwelint: It's handled slightly differently in different formats, e.g. record format(s) for example any integral type, except `long` if it's very large, will fit in one "block" i.e. 8B. Block format tho has dynamic compression based on integral value size, not its type (even works for negative values). So saying that storing primitive integrals as 64-bit long integers instead of their respective type takes up more space is mostly incorrect.

This is what I meant yesterday on the call that storing single primitive integrals doesn't increase store size, whereas doing this to arrays will.

mauvo: @tinwelint is this what you're saying?

1) On Record Format: a byte, short, integer, and long, will all be stored in an 8B block regardless.
2) On Block Format: the number 1 will be stored in a little space as possible regardless of its typing as byte, short, integer or long.

The above I think help explain the thinking behind normalize types. So we should drop the part about saving space, and simply note it simplifies things for Cypher.

tinwelint: Yes you're right I think that's good to not involve the space aspect in the description since at least for integral types the only outlier is a negative long in record format(s). Perhaps there's some difference for float/double, but to venture into it would probably just complicate matters.

renetapopova:
Closing this one - changes copied to https://github.com/neo4j/docs-operations/pull/1221

PR #1240, opened 2023-12-04
Update plugins.adoc in Kubernetes

davidlrosenblum:
Removed the section where we installed apoc with the env ["apoc"].  This is not needed if we are just pointing to the file in labs.


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1240.surge.sh

PR #1244, opened 2023-12-07
Deploy an analytics cluster using the Neo4j Helm chart

renetapopova:


modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
```diff
@@ -0,0 +1,168 @@
+:description: How to deploy a Neo4j cluster to a cloud or a local Kubernetes cluster using Neo4j Helm chart.
+[role=enterprise-edition]
+[[quick-start-analytic-cluster]]
+= Quickstart: Deploy a Neo4j cluster for analytic queries
+
+_The feature is available in the Neo4j Helm chart from version 5.14._
+
+This quickstart shows how to configure and deploy a special Neo4j cluster without fault tolerance to support analytic queries.
```
harshitsinghvi22, line 8: use of words like "without fault tolerance" is fine but we should mention that this scenario is specifically and only for "1 primary and N secondary" ... without fault tolerance does imply that but not everyone will understand that hence if we can mention its "1 primary and N secondary" that would be great

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
```diff
@@ -0,0 +1,168 @@
...
+Information on using Neo4j’s graph data science library in a cluster can be found in the link:https://neo4j.com/docs/graph-data-science/current/[Neo4j Graph Data Science library documentation].
+
+The cluster is deployed to a cloud or a local Kubernetes cluster using the Neo4j Helm chart.
+Because the GDS library does not support fault tolerance, you can deploy a cluster with one primary server for the transaction workload and an N number of secondary servers for the GDS analytics.
```
harshitsinghvi22, line 12: GDS library does not support fault tolerance ? are you sure ? i don't think so , you can install GDS in a cluster of 5 members where we have 3 primary and 2 secondary this makes it fault tolerant.

So , i guess this statement needs to be removed.

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
```diff
@@ -0,0 +1,168 @@
...
+services:
+  neo4j:
+    enabled: false
+  internals:
+    enabled: true
+
+# Enable the analytics cluster and set the type to primary:
+analytics:
+  enabled: true
+  type: primary
```
harshitsinghvi22, line 60: this is wrong , the correct syntax is this. Please refer to the demo video again.
```
analytics:
  enabled: true
  type: 
    name: primary
```

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
```diff
@@ -0,0 +1,168 @@
...
+services:
+  neo4j:
+    enabled: false
+  internals:
+    enabled: true
+
+# Enable the analytics cluster and set the type to secondary:
+analytics:
+  enabled: true
+  type: secondary
```
harshitsinghvi22, line 110: same thing here

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
```diff
@@ -0,0 +1,168 @@
...
+----
+. When deployed, copy the external IP of the LoadBalancer to access Neo4j from an application outside the Kubernetes cluster.
+For more information, see xref:kubernetes/accessing-neo4j.adoc#access-outside-k8s[Applications accessing Neo4j from outside Kubernetes].
+. In a web browser, open the Neo4j Browser at _http://EXTERNAL_IP:7474/browser_ and log in using the password you have configured in your values YAML files.
+. Run the following Cypher query to verify that the cluster is deployed and running:
++
+[source, cypher]
+----
+SHOW SERVERS
+----
```
harshitsinghvi22, line 154: this is missing another CYPHER COMMAND. 'SHOW DATABASES'
it would be good to put a screenshot here and explain what exactly in the output of the above command ensures the setup is complete.

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
```diff
@@ -0,0 +1,168 @@
...
+. Run the following Cypher function to verify that the GDS library is installed:
++
```
harshitsinghvi22, line 156: the cypher queries for gds might not work through the browser since the query might get executed on the primary instance and since that does not have GDS installed it will lead in wrong output.

The gds queries as mentioned in the demo video are to be executed through cypher shell only. Please refer to the demo video again.

harshitsinghvi22: You can execute this through browser but then you have to login through the browser via bolt protocol on the specific gds instance which might make it complicated hence its better to mention the cypher shell executed in the  gds pod itself

harshitsinghvi22 (changes requested):

modules/ROOT/pages/kubernetes/index.adoc
```diff
@@ -15,6 +15,7 @@ This chapter describes the following:
 * xref:kubernetes/helm-charts-setup.adoc[Configure the Neo4j Helm chart repository] -- Configure the Neo4j Helm chart repository and check for the available charts.
 * xref:kubernetes/quickstart-standalone/index.adoc[Quickstart: Deploy a standalone instance] -- Deploy a Neo4j standalone instance to a cloud (GKE, AWS, AKS) or a local (via Docker Desktop for macOS) Kubernetes cluster.
 * xref:kubernetes/quickstart-cluster/index.adoc[Quickstart: Deploy a cluster] -- Deploy a Neo4j cluster to a cloud (GKE, AWS, AKS) Kubernetes cluster.
+* xref:kubernetes/quickstart-analytics-cluster.adoc[Quickstart: Deploy a Neo4j cluster for analytic queries] -- Deploy an analytics Neo4j cluster to a local or a cloud (GKE, AWS, AKS) Kubernetes cluster.
```
harshitsinghvi22, line 18: it would be good to be specific in the title "Deploy a Neo4j Cluster for analytics (1 primary + N Secondary)" since N primary and N secondary was already possible

harshitsinghvi22 (changes requested):

modules/ROOT/pages/kubernetes/index.adoc
renetapopova: I thought that we could expand this page similar to how this topic is presented in the [Clustering chapter ](https://neo4j.com/docs/operations-manual/current/clustering/setup/analytics-cluster/), and have two sections an analytics cluster with fault-tolerance and without.

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
renetapopova: This decision was taken due to my intention above.

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
renetapopova: If that's not true, then the docs are not correct:
"The analytics cluster can be set up in two ways, with fault tolerance or without fault tolerance. Bear in mind that the GDS library does not support fault tolerance and therefore GDS should only be deployed on servers configured for analytic queries, as described below."

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
renetapopova: I'll watch the demo again. I thought that was what you were showing but I might had been wrong,

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
renetapopova: It's not a good practice to use screenshots where we can add a code snippet. Could you provide me with the results of this commands?

renetapopova:
@harshitsinghvi22, could you please review this PR?

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
```diff
@@ -0,0 +1,180 @@
...
+[source, bash]
+----
+helm install lb neo4j/neo4j-load-balancer --set neo4j.name="analytics-cluster"
+----
+. When deployed, copy the `EXTERNAL_IP` of the LoadBalancer service.
+For more information, see xref:kubernetes/quickstart-cluster/access-outside-k8s.adoc[Access the Neo4j cluster from outside Kubernetes].
+. In a web browser, open the Neo4j Browser at _http://EXTERNAL_IP:7474/browser_ and log in using the password you have configured in your values YAML files.
+. Run the `SHOW SERVERS` command to verify that the cluster is deployed and running.
+. Run the `SHOW DATABASES` command to verify that primary server is of type `standalone` and the secondary servers are of type `gds`.
```
harshitsinghvi22, line 180: shouldn't this be the other way around ? standalone is of type primary and gds of type secondary

harshitsinghvi22 (changes requested):

modules/ROOT/pages/kubernetes/quickstart-analytics-cluster.adoc
renetapopova: of course

harshitsinghvi22 (approved):

PR #1260, opened 2023-12-18
Moving the password page to auth

lidiazuin:


lidiazuin:
@renetapopova I was wondering if https://neo4j.com/docs/operations-manual/5/configuration/set-initial-password/ should also be moved to the authentication and authorization section, but I think no?

neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1260.surge.sh

modules/ROOT/content-nav.adoc
```diff
@@ -178,6 +177,7 @@
 ** Integration with auth systems
 *** xref:authentication-authorization/ldap-integration.adoc[]
 *** xref:authentication-authorization/sso-integration.adoc[]
+** xref:authentication-authorization/password-and-user-recovery.adoc[]
```
renetapopova, line 180: I think we should experiment a bit and see if this page can be incorporated into the **Manage users** page, or else we could add it after ** xref:authentication-authorization/manage-roles.adoc[].

modules/ROOT/pages/authentication-authorization/password-and-user-recovery.adoc
```diff
@@ -1,4 +1,5 @@
 :description: This page describes how to reset a password to recover a user's access when their password is lost. It specifically focuses on how to recover an admin user if all the admin users have been unassigned the admin role, and how to recreate the built-in admin role if it has been dropped.
+:page-aliases: configuration/password-and-user-recovery.adoc
```
renetapopova, line 2: I prefer to move content instead of using aliases because, after some time, we might end up with many aliases and find it challenging to find the actual content.

lidiazuin: No, but it was moved. The alias is just in case there is another page where the old url is used (which there is), so it's redirected to the new path, so avoiding an error 404.

modules/ROOT/content-nav.adoc
lidiazuin: I think it's a good idea to incorporate the content to the Manage users page. Should I commit this change here or open another PR?

modules/ROOT/pages/authentication-authorization/password-and-user-recovery.adoc
renetapopova: Ah, ok. I didn't know that page aliases can be used instead of redirects.

modules/ROOT/content-nav.adoc
renetapopova: Hmm, let's keep it in one PR so that we see what we are doing.

renetapopova:
@lidiazuin, OK, let's keep it a separate page on the same level but under **Manage roles**. We can update the title to be consistent with the rest of the pages and the content on the page, for example, **Recover admin user and password**. This page also needs to be updated as it shows how to do it in 4.4 but not in 5.x.

modules/ROOT/pages/authentication-authorization/password-and-user-recovery.adoc
```diff
@@ -1,6 +1,7 @@
...
 [[password-and-user-recovery]]
-= Password and user recovery
+=  Recover user and password
```
renetapopova, line 4:
```suggestion
=  Recover admin user and password
```

renetapopova (changes requested):
I left some comments.

modules/ROOT/pages/authentication-authorization/password-and-user-recovery.adoc
lidiazuin: Ah yes, but the links are still functioning because of the alias. In any case, it's good to have this updated in the page itself. Thanks for flagging it!

PR #1317, opened 2024-01-09
Fix duplicated entry for `page_cache.page_cancelled_faults` metric

NataliaIvakina:
There is a duplicated entry for that metric, with the only difference being one has the description spelled `canceled` and the other one `cancelled`.

I merged both into the first entry just because because that matches the order of the metrics in the code-base.


neo-technology-commit-status-publisher:

You can view the updated docs at https://neo4j-docs-operations-1317.surge.sh

PR #1336, opened 2024-01-18
Tag deallocated state with version

gfx54b:
The operations manual refers to the `Deallocated` state as it was always available. Actually it is a rather new invention, so it is worth pointing out that to the reader.


neo-technology-commit-status-publisher:

Check out your changes at https://neo4j-docs-operations-1336.surge.sh

PR #1348, opened 2024-01-22
Update scaling.adoc

wawrzek:
JVM Heap initial memory set twice. That's probably a mistake, should be max_size in the later config lines.


renetapopova:
Good catch, @wawrzek! Thanks. Fixed in #1371.

PR #1403, opened 2024-02-02
updated mentions of ubi8 docker images to ubi9

jennyowen:
In 5.17.0 we will start deprecating ubi8 images in favour of ubi9 ones. This updates the documentation so that we aren't telling users to start the deprecated images.

This should only be merged into the public documentation after the 5.17 release though, otherwise the docker commands won't work. I'm not sure what needs doing to ensure it doesn't get published early?

### DIFF
--- a/modules/ROOT/pages/docker/introduction.adoc
+++ b/modules/ROOT/pages/docker/introduction.adoc
@@ -35,7 +35,7 @@ ____
 
 === Base operating system
 
-The Neo4j image is available with either `debian:bullseye-slim` or `redhat/ubi8-minimal:latest` as the base image.
+The Neo4j image is available with either `debian:bullseye-slim` or `redhat/ubi9-minimal:latest` as the base image.
 The default is `debian:bullseye-slim`.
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1403.surge.sh

Updated pages:
- [Getting started with Neo4j in Docker](https://neo4j-docs-operations-1403.surge.sh/operations-manual/5/docker/introduction)

renetapopova:
> In 5.17.0 we will start deprecating ubi8 images in favour of ubi9 ones. This updates the documentation so that we aren't telling users to start the deprecated images.
> 
> This should only be merged into the public documentation after the 5.17 release though, otherwise the docker commands won't work. I'm not sure what needs doing to ensure it doesn't get published early?

Thanks, @jennyowen, I marked the PR with v5.17, which means to be published as part of the 5.17 release.

jennyowen:
Thanks @renetapopova !

PR #1405, opened 2024-02-05
Adding new page alias after several broken links in different guides

lidiazuin:
Even though we have the `manage-databases/introduction.adoc` page alias, in several guides (Go, JavaScript, Bloom etc) the link `manage-databases/introduction.adoc#manage-databases-default` is being used instead and the `page-aliases` component doesn't seem to map that. We could change the link in every guide it appears or use this PR to add a new alias to fix the 404 error.

### DIFF
--- a/modules/ROOT/pages/database-administration/index.adoc
+++ b/modules/ROOT/pages/database-administration/index.adoc
@@ -1,5 +1,5 @@
 :description: This page gives an introduction to database administration using Neo4j.
-:page-aliases: manage-databases/introduction.adoc, manage-databases/index.adoc
+:page-aliases: manage-databases/introduction.adoc, manage-databases/index.adoc, manage-databases/introduction.adoc#manage-databases-default
 [database-administration]
 = Database administration
 


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1405.surge.sh

Updated pages:
- [Database administration](https://neo4j-docs-operations-1405.surge.sh/operations-manual/5/database-administration)

modules/ROOT/pages/database-administration/index.adoc
```diff
@@ -1,5 +1,5 @@
 :description: This page gives an introduction to database administration using Neo4j.
-:page-aliases: manage-databases/introduction.adoc, manage-databases/index.adoc
+:page-aliases: manage-databases/introduction.adoc, manage-databases/index.adoc, manage-databases/introduction.adoc#manage-databases-default
```
renetapopova, line 2: I tried with anchors, but it didn't work. Have you tested this one?

renetapopova: I would suggest that we update the links in all those manuals. It's a cleaner solution.

renetapopova: Actually, it's a redirect thing. We need to redirect the `manage-databases/introduction` to `database-administration`.

renetapopova: Let me know if you want me to add it or you'll do it.

lidiazuin: Yes but we are already redirecting `manage-databases/introduction` to `database-administration`, the thing is that in Bloom and other drivers manuals the link includes an anchor. I tried the redirect with the anchor locally and it worked, but I agree that changing in the other manuals might be a cleaner option!

renetapopova: No, we don't. There is no redirect, just a page-aliase.  Page-aliases do not span across manuals.

renetapopova: We need to create a redirect anyway, as every page-alias must be backed by a redirect. Then, those links will work as they will be redirected to the proper page.

lidiazuin: How do you want to create the redirect, then? I thought we were supposed to use page-aliases

renetapopova: The redirects are created in the manual-modeling-antora repo - see the details here: https://github.com/neo-technology/neo4j-manual-modeling-antora/wiki/Redirects

lidiazuin: I'll close this PR then and work on the redirects. Thanks!

PR #1421, opened 2024-02-09
Update the store  format page

renetapopova:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1421.surge.sh

Updated pages:
- [Store formats](https://neo4j-docs-operations-1421.surge.sh/operations-manual/5/database-internals/store-formats)

modules/ROOT/pages/database-internals/store-formats.adoc
```diff
@@ -11,6 +11,9 @@ The available formats are:
 
 Neo4j's newest store format is `block`.
 It has the best performance and supports the highest limits.
+For example, compared to the existing record format, the `block` format has roughly 40% better performance when the whole graph is in memory.
+That lead increases to 70% when only a third of the graph can fit in memory.
```
Roxling, line 15: I don't think we should have things like performance numbers in official documentation. It is not a guarantee and is highly data dependent, some users may see much better results, and some no increase at all

modules/ROOT/pages/database-internals/store-formats.adoc
```diff
@@ -25,14 +28,46 @@ The `high-limit` format allows more nodes and relationships to be stored than `s
 
 You can either set the store format when creating a new database or change the store format of an existing database.
 
+The default store format of new databases is `aligned` and is controlled by the xref:configuration/configuration-settings.adoc#config_db.format[`db.format`] configuration setting in the _neo4j.conf_ file.
```
Roxling, line 31: Why move this text up from the **Creating new databases** where it was? That setting controls format for **_new_** databases only

Roxling, line 31: Why mention the default value when we link to the setting. I don't think we do that for any other setting in the documentation. Also, its easy to forget to update this page when we update the default value in the future

modules/ROOT/pages/database-internals/store-formats.adoc
```diff
@@ -25,14 +28,46 @@ The `high-limit` format allows more nodes and relationships to be stored than `s
...
-The default store format of new databases is controlled by the xref:configuration/configuration-settings.adoc#config_db.format[`db.format`] configuration setting in the _neo4j.conf_ file.
-Or the store format can be passed as an argument to the tool creating the database, e.g. xref:tools/neo4j-admin/neo4j-admin-import.adoc#import-tool-full[`neo4j-admin database import`] or xref:backup-restore/copy-database.adoc[`neo4j-admin database copy`] commands.
+There are several ways to create a new database in a specific store format:
+
+* Specify the store format when creating a new database using the `CREATE DATABASE` Cypher statement.
+For example:
++
+[source,cypher]
+----
+CREATE DATABASE blockdb OPTIONS {storeFormat: 'block'}
```
Roxling, line 42: This is still not supported official syntax https://neo4j.com/docs/operations-manual/current/database-administration/standard-databases/manage-databases/#manage-databases-create-database-options
It should not be in the official documentation

modules/ROOT/pages/database-internals/store-formats.adoc
```diff
@@ -42,6 +77,26 @@ Changing the store format requires that:
...
+SHOW DATABASES YIELD name, store
+----
+
+.Result
+[role="queryresult"]
+----
++-----------------------------------+
+| name      | store                 |
++-----------------------------------+
+| "blockdb" | "block-block-2b.1"    |
```
Roxling, line 95: This is the beta format which isn't supported since 5.14 and I don't think we mention here. The correct format is `block-block-1.1`

Roxling (changes requested):
**In my opinion** most of these changes should not be in the official documentation, as they add no value. There is a reason these things were put in a blog post. It is not the purpose of this page to get people over to block format and the examples provided should live in the documentation of the individual tools, as it was.

modules/ROOT/pages/database-internals/store-formats.adoc
renetapopova: You are right, but my idea is that by the config you set a default format for all new databases. And the heading below is about a creating a single database on a specific format.

modules/ROOT/pages/database-internals/store-formats.adoc
renetapopova: So, it seems that David Pond's blog post is not correct, which by the way is about the beta version of the block format.

modules/ROOT/pages/database-internals/store-formats.adoc
renetapopova: My idea was to tell the users what the default format is for all new databases and how to change it if they want.

modules/ROOT/pages/database-internals/store-formats.adoc
renetapopova: Users want more specific information about the differences and which format to use and when, but I agree that these percentages may change.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1421.surge.sh

Updated pages:
- [Store formats](https://neo4j-docs-operations-1421.surge.sh/operations-manual/5/database-internals/store-formats)
- [Import](https://neo4j-docs-operations-1421.surge.sh/operations-manual/5/tools/neo4j-admin/neo4j-admin-import)

modules/ROOT/pages/database-internals/store-formats.adoc
```diff
@@ -28,29 +26,31 @@ The `high-limit` format allows more nodes and relationships to be stored than `s
 
 You can either set the store format when creating a new database or change the store format of an existing database.
 
-The default store format of new databases is `aligned` and is controlled by the xref:configuration/configuration-settings.adoc#config_db.format[`db.format`] configuration setting in the _neo4j.conf_ file.
-
 === Creating new databases
 
-There are several ways to create a new database in a specific store format:
+The default store format for all new databases is `aligned`.
```
Roxling, line 31: I still don't think this adds any value. But if you really want it, we can at least make a note that this may change in future versions

modules/ROOT/pages/tools/neo4j-admin/neo4j-admin-import.adoc
```diff
@@ -409,7 +409,7 @@ label:enterprise-only[]
 
 [WARNING]
 ====
-Incremental import is not supported for `block` format yet.
+Incremental import does not support `block` format.
```
Roxling, line 412: This functionality is currently being built. We should not forget to remove this note once it is supported

Roxling (approved):

PR #1440, opened 2024-02-26
Adding clarification that sub-admin roles should not be able to change their own privileges

phil198:
This was noticed in a recent internal pentest of RBAC.

We need to clarify that it is necessary to explicitly deny the ability to change your own privileges if you copy the admin role (otherwise the new role is essentially unconstrained).


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1440.surge.sh

Updated pages:
- [Fine-grained access control (example)](https://neo4j-docs-operations-1440.surge.sh/operations-manual/5/tutorial/access-control)

Hunterness (commented):
I know we assigned Lasse to this but looks good to me

renetapopova:
@phil198, Which version should this go into, 5.18 or current?

PR #1450, opened 2024-02-27
Adding clarification that sub-admin roles should not be able to change their own privileges 

renetapopova:
Cherry-picked from #1440 

This was noticed in a recent internal pentest of RBAC.

We need to clarify that it is necessary to explicitly deny the ability to change your own privileges if you copy the admin role (otherwise the new role is essentially unconstrained).


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1450.surge.sh

Updated pages:
- [Fine-grained access control (example)](https://neo4j-docs-operations-1450.surge.sh/operations-manual/5/tutorial/access-control)

PR #1459, opened 2024-02-29
update the reverse-proxy yaml to support privilege and access control settings for a Container

renetapopova:


renetapopova:
approved on slack

PR #1478, opened 2024-03-08
Document db.create.setRelationshipVectorProperty procedure

renetapopova:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1478.surge.sh

Updated pages:
- [Procedures](https://neo4j-docs-operations-1478.surge.sh/operations-manual/5/reference/procedures)

modules/ROOT/pages/reference/procedures.adoc
```diff
@@ -972,6 +977,19 @@ m| db.create.setNodeVectorProperty(node :: NODE, key :: STRING, vector :: LIST<F
...
+[role=label--new-5.18 label--beta]
+[[procedure_db.create.setRelationshipVectorProperty]]
+.db.create.setRelationshipVectorProperty()
+[cols="<15s,<85"]
+|===
+| Description
+a|
+Set a vector property on a given relationship in a more space efficient representation than Cypher's link:{neo4j-docs-base-uri}/cypher-manual/{page-version}/clauses/set#set-set-a-property[`SET`].
+| Signature
+m| db.create.setRelationshipVectorProperty(relationship :: RELATIONSHIP, key :: STRING, vector :: ANY)
```
parnmatt, line 989: `ANY` is a limitation of the procedure framework, and how it currently works to try and automatically figure out the type … which is only used for that test against the JSON file.

It's real signature would be `LIST<INTEGER | FLOAT>`.

```suggestion
m| db.create.setRelationshipVectorProperty(relationship :: RELATIONSHIP, key :: STRING, vector :: LIST<INTEGER | FLOAT>)
```

and _now_ same above for `setNodeVectorProperty`, (and `setVectorProperty`
I'm just confirming with @nilsceberg if it is an additive change, or technically always was that and thus a correction to the docs.

parnmatt: Nils has just confirmed with 5.16 that it worked with a `LIST<INTEGER>`; therefore, updating `setNodeVectorProperty` and `setVectorProperty` to `LIST<INTEGER | FLOAT>` is a docs correction and shouldn't required noting it as a signature change.

parnmatt:
Should we also document `db.index.vector.queryRelationships` here as well, or would it be best to do so in another PR?
_(thus also `db.index.vector.queryNodes` with `LIST<FLOAT>` to `LIST<INTEGER | FLOAT>`)_

modules/ROOT/pages/reference/procedures.adoc
renetapopova: Hmm, this is from the codebase.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1478.surge.sh

Updated pages:
- [Procedures](https://neo4j-docs-operations-1478.surge.sh/operations-manual/5/reference/procedures)

modules/ROOT/pages/reference/procedures.adoc
parnmatt: Indeed, I believe you're specifically referring to `procedures.json`, which is only used in procedure and function acceptance tests from cypher ops. I guess docs use this for reference too in that case?

---

For reference and clarity,
The procedure framework unfortunately sees almost anything that uses one of our internal types as `ANY`.
Even if it was a `FloatingPointArray` which _is_ `LIST<FLOAT>`, it would be `ANY` in that type detection.

Semantically, and through enforcement of manual type checking, it should be `LIST<INTEGER | FLOAT>` here.
We are using the internal types for performance reasons, and this is an implementation detail. 

If we were writing this in scala where some of their procedures/functions are, I believe the signature checker would use the static one that is provided, and thus correctly pick up it was a `LIST<INTEGER | FLOAT>` but I could be wrong.

Either way, the `ANY` in that JSON file are to satisfy those tests.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1478.surge.sh

Updated pages:
- [Procedures](https://neo4j-docs-operations-1478.surge.sh/operations-manual/5/reference/procedures)

modules/ROOT/pages/reference/procedures.adoc
renetapopova: Yes, we use the Cypher command `SHOW PROCEDURES` to validate the available procedures, their description, signature, and mode. I can see that the `genai.vector.encodeBatch` is missing in the json now. What happened to it?

parnmatt: > Yes, we use the Cypher command SHOW PROCEDURES to validate the available procedures, their description, signature, and mode.

Right! That might be an issue. Of course, that JSON file is specifically there testing the output for `SHOW PROCEDURES`. So the signature is _there_ … It would be _best_ if it were documented as `LIST<INTEGER | FLOAT>` as that is the semantic (and enforced) type for that parameter.
The displaying of `ANY` could be considered a 'bug'. However, if you need to be consistent with `SHOW PROCEDURES` then I guess you may have to be `ANY` … in which case `queryNodes`, `queryRelationships`, `setNodeVectorProperty`, `setRelationshipProperty`, and `setVectorProperty` _all_ would have to have `ANY` now rather than `LIST<INTEGER | FLOAT>` (or `LIST<FLOAT>` as some of them currently are)

parnmatt: > I can see that the genai.vector.encodeBatch is missing in the json now. What happened to it?

It never was in the JSON. It is not a built-in 'Neo4j' procedure, it's an external procedure from a plugin (like APOC or GDS procedures).

nilsceberg: Going back from `ANY` to `LIST<INTEGER | FLOAT>` in the _documentation_ (if we are able to that in the future) seems like more of a breaking change than treating the `SHOW PROCEDURES` output as a bug. :S

renetapopova: > > I can see that the genai.vector.encodeBatch is missing in the json now. What happened to it?
> 
> It never was in the JSON. It is not a built-in 'Neo4j' procedure, it's an external procedure from a plugin (like APOC or GDS procedures).

Ah, yes, but I guess it will appear in the output when you install the plugin.

renetapopova: > Going back from `ANY` to `LIST<INTEGER | FLOAT>` in the _documentation_ (if we are able to that in the future) seems like more of a breaking change than treating the `SHOW PROCEDURES` output as a bug. :S

So, are you suggesting to ignore the tests and just add `LIST<INTEGER | FLOAT>` instead of `ANY` even though the users will see a different signature when they run `SHOW PROCEDURES`?

modules/ROOT/pages/reference/procedures.adoc
```diff
@@ -1155,6 +1178,20 @@ m| db.index.vector.queryNodes(indexName :: STRING, numberOfNearestNeighbours ::
...
 |===
 
+[role=label--new-5.18 label--beta]
+[[procedure_db.index.vector.queryRelationships]]
+.db.index.vector.queryRelationships()
+[cols="<15s,<85"]
+|===
+| Description
+a|
+Query the given relationship vector index.\nReturns requested number of nearest neighbors to the provided query vector, and their similarity score to that query vector, based on the configured similarity function for the index. The similarity score is a value between `[0, 1]`; where `0` indicates least similar, `1` most similar.
```
NataliaIvakina, line 1188:
```suggestion
Query the given relationship vector index.
Returns requested number of nearest neighbors to the provided query vector, and their similarity score to that query vector, based on the configured similarity function for the index. The similarity score is a value between `[0, 1]`; where `0` indicates least similar, `1` most similar.
```

modules/ROOT/pages/reference/procedures.adoc
nilsceberg: I think we should at least consider that, yes. Perhaps we could note the discrepancy in the docs?

renetapopova: I am not sure what we should do in such cases. The documentation reflects what's in the output of SHOW PROCEDURES because that's what users see. If the output shows ANY, the docs also show ANY. And as per the docs test, everything seems fine.

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1478.surge.sh

Updated pages:
- [Procedures](https://neo4j-docs-operations-1478.surge.sh/operations-manual/5/reference/procedures)

modules/ROOT/pages/reference/procedures.adoc
nilsceberg: > And as per the docs test, everything seems fine.

Wait, are the tests passing for `queryNodes`? It currently seems to say `query :: LIST<FLOAT>` in the docs but that has also recently changed to `ANY` in the `SHOW PROCEDURES` output, unless I'm mistaken :thinking:

renetapopova: > I think we should at least consider that, yes. Perhaps we could note the discrepancy in the docs?

What would be a good note for that? Could you provide us with some content?

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1478.surge.sh

Updated pages:
- [Procedures](https://neo4j-docs-operations-1478.surge.sh/operations-manual/5/reference/procedures)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1478.surge.sh

Updated pages:
- [Procedures](https://neo4j-docs-operations-1478.surge.sh/operations-manual/5/reference/procedures)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1478.surge.sh

Updated pages:
- [Procedures](https://neo4j-docs-operations-1478.surge.sh/operations-manual/5/reference/procedures)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1478.surge.sh

Updated pages:
- [Procedures](https://neo4j-docs-operations-1478.surge.sh/operations-manual/5/reference/procedures)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1478.surge.sh

Updated pages:
- [Procedures](https://neo4j-docs-operations-1478.surge.sh/operations-manual/5/reference/procedures)

NataliaIvakina (approved):

PR #1486, opened 2024-03-14
Add Store Format link to `--format` parameter entry

brunobuss:
Since we don't list anywhere the list of available store formats
(since these can change based on the version and edition of the DBMS),
we can point to our Store Formats documentation to allow users figure
out what are the options they could use there.


renetapopova (approved):
Looks good. Thanks, @brunobuss!

PR #1487, opened 2024-03-14
Cherrypick into 5.x: Add Store Format link to `--format` parameter entry

brunobuss:
Since we don't list anywhere the list of available store formats (since these can change based on the version and edition of the DBMS), we can point to our Store Formats documentation to allow users figure out what are the options they could use there.

This is a cherry pick of #1486.


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1487.surge.sh

Updated pages:
- [Import](https://neo4j-docs-operations-1487.surge.sh/operations-manual/5/tools/neo4j-admin/neo4j-admin-import)

PR #1571, opened 2024-04-22
Use unicode on CSV file delimiters example

brunobuss:
Change the `--array-delimiter` example to be set using the code point for the character `|` (the same used before, so we don't need to change the CSV content).

This was based on a user feedback, that it was not clear from the examples that they could have used (and how they could have used) unicode code points to set the delimiter to arbitrary characters, including non-printable ones.

See https://github.com/neo4j/neo4j/issues/13445.

### DIFF
--- a/modules/ROOT/pages/tutorial/neo4j-admin-import.adoc
+++ b/modules/ROOT/pages/tutorial/neo4j-admin-import.adoc
@@ -174,7 +174,7 @@ The details of a CSV file header format can be found at xref:tools/neo4j-admin/n
 The following CSV files have:
 
 * `--delimiter=";"`
-* `--array-delimiter="|"`
+* `--array-delimiter="U+007C"` (`0x007C` is the Unicode code point for the character `|`)
 * `--quote="'"`
 
 .movies2.csv
@@ -219,7 +219,7 @@ The call to `neo4j-admin database import` would look like this:
 .shell
 [source]
 ----
-bin/neo4j-admin database import full --delimiter=";" --array-delimiter="|" --quote="'" --nodes=import/movies2.csv --nodes=import/actors2.csv --relationships=import/roles2.csv neo4j
+bin/neo4j-admin database import full --delimiter=";" --array-delimiter="U+007C" --quote="'" --nodes=import/movies2.csv --nodes=import/actors2.csv --relationships=import/roles2.csv neo4j
 ----
 
 


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1571.surge.sh

Updated pages:
- [Neo4j-admin import](https://neo4j-docs-operations-1571.surge.sh/operations-manual/5/tutorial/neo4j-admin-import)

modules/ROOT/pages/tutorial/neo4j-admin-import.adoc
```diff
@@ -174,7 +174,7 @@ The details of a CSV file header format can be found at xref:tools/neo4j-admin/n
 The following CSV files have:
 
 * `--delimiter=";"`
-* `--array-delimiter="|"`
+* `--array-delimiter="U+007C"` (`0x007C` is the Unicode code point for the character `|`)
```
parnmatt, line 177: you've mixed conventions here using `U+` and `0x`
also, have you tested this works with that conversion? Doesn't java use the `\u` syntax  for defining characters in unicode thus  `\u007c` ?

brunobuss: > you've mixed conventions here using U+ and 0x

That was intentional, because I had the idea that places refers to unicode code points as 4 hex digit, so the `0x` was clear on my mind. But the actual code point terminology (and the one we support) uses the `U+` prefix.

Do you think it would be better to have both as `U+`?

> also, have you tested this works with that conversion? Doesn't java use the \u syntax for defining characters in unicode thus \u007c ?

You can see on `ImportCommand.java` how our delimiter flags uses the `EscapedCharacterConverter`, which in turn is the `CharacterConverter` that supports the following conventions for unicode expressed as hex: `\u12AB`, `\U12AB`, `u+12AB`, `U+12AB`.

Even though the `\u` is indeed the java way to represent unicode, I personally like more the `U+` as this aligns with unicode documentation and "feels like" language independent :)

parnmatt: > Do you think it would be better to have both as `U+`?

Yes, I think we should be consistent. The `U+` convention is the standard way unicode code points are referred to; however, they are not always supported to be typed in a coding language.

So one can argue that the comment to the side _should_ be using the `U+` anyway, and the example in code preceding it, using the supported programmatic convention supported by tool/language/etc.

If the `U+` conversion happens to work with out tool that's awesome! Then we can use that.

_(It wouldn't hurt to note the other working conventions, such as `\u`, `\x`, `0x`, etc._ if they work, _as an aside or footnote, but I don't think it's massively needed)_

brunobuss: There we go: https://github.com/neo4j/docs-operations/pull/1575

I'll cherrypick both of them together into the 5.x branch on https://github.com/neo4j/docs-operations/pull/1573

PR #1573, opened 2024-04-22
Use unicode on CSV file delimiters example

brunobuss:
Change the `--array-delimiter` example to be set using the code point for the character `|` (the same used before, so we don't need to change the CSV content).

This was based on a user feedback, that it was not clear from the examples that they could have used (and how they could have used) unicode code points to set the delimiter to arbitrary characters, including non-printable ones.

Cherry-picked from https://github.com/neo4j/docs-operations/pull/1571.


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1573.surge.sh

Updated pages:
- [Neo4j-admin import](https://neo4j-docs-operations-1573.surge.sh/operations-manual/5/tutorial/neo4j-admin-import)

renetapopova (approved):

PR #1575, opened 2024-04-23
Don't use `0x` prefix to reference the unicode code point

brunobuss:
This is to avoid confusion, since we *don't* support passing the delimiter with the `0x` prefix.


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1575.surge.sh

Updated pages:
- [Neo4j-admin import](https://neo4j-docs-operations-1575.surge.sh/operations-manual/5/tutorial/neo4j-admin-import)

renetapopova (approved):

PR #1746, opened 2024-07-26
updated ssl+netty compatibility table for 5x

jennyowen:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1746.surge.sh

Updated pages:
- [SSL framework](https://neo4j-docs-operations-1746.surge.sh/operations-manual/5/security/ssl-framework)

modules/ROOT/pages/security/ssl-framework.adoc
```diff
@@ -50,17 +50,101 @@ The following table includes detailed information about supported Neo4j versions
 | tcnative version
 | Direct link
 
+| 5.22
```
NataliaIvakina, line 53: everything looks good. But I think there is no need to show Netty versions for Neo4j 5.12-latest, since starting from 5.12 you can find netty and tcnative files in the _lib_ directory

NataliaIvakina: thus my suggestion is to remove info for  Neo4j 5.12-5.22 from this table

jennyowen: the netty jars being included was a bug and it got fixed in 5.20 or 5.21 (I forget which).
Those bundled versions were the boringSSL variant. From 5.23 onwards we will (hopefully) bundle the dynamically linked tcnative version.

Reducing the table size would be nice but I think overall, just listing every compatibility is maybe the more readable option.

That being said, the netty version doesn't actually matter, it's the `tcnative` version that is relevant for SSL, and that changes way less frequently.

NataliaIvakina (approved):

NataliaIvakina:
@jennyowen, hi! Thank you for your explanation! I think this PR can be merged. I can take care of cherry-picking it into the publish branch.

jennyowen:
@NataliaIvakina ok will do, thanks!

PR #1815, opened 2024-09-17
Add a note about clusters with 2 members

gfx54b:
Recent question in slack pointed out that we describe how to set up a 3 cluster (as that should be the normal), but we don't mention anywhere that in case of a 2 member cluster an extra setting is needed.

PR #1817, opened 2024-09-18
Add a note about clusters with 2 members (#1815)

NataliaIvakina:
Recent question in slack pointed out that we describe how to set up a 3 cluster (as that should be the normal), but we don't mention anywhere that in case of a 2 member cluster an extra setting is needed.

---------


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1817.surge.sh

Updated pages:
- [Deploy a basic cluster](https://neo4j-docs-operations-1817.surge.sh/operations-manual/5/clustering/setup/deploy)

PR #1866, opened 2024-10-15
Removed message about no TLS/SSL support for Minio.

mauvo:
We think it probably works, we just haven't tested it. So let's not say one way or the other just yet.


mauvo:
See Slack: https://neo4j.slack.com/archives/C03UYP33BPC/p1729004916942719

PR #1862, opened 2024-10-15
remove before version 4.4

fiquick:
as per ticket https://trello.com/c/uUkM9CM5/350-document-change-console-drag-and-drop-import-to-support-version-44-or-above

PR #1875, opened 2024-10-17
delete push to cloud page

fiquick:
See also https://github.com/neo4j/docs-operations/pull/1862


NataliaIvakina:
What happened to the command `push-to-cloud` in 4.3? Is it deprecated, not supported? See https://neo4j.com/docs/operations-manual/4.3/tools/

renetapopova:
4.3 is no longer a supported version, but we can't just delete a page because whatever is in the manual was supported when this version was supported. I think a better option would be to hide all unsupported versions from the version picker. I have created a card for that https://trello.com/c/G23oIBIg/56-hide-all-versions-that-we-dont-support-from-the-version-picker-in-the-operations-manual

NataliaIvakina:
> 4.3 is no longer a supported version, but we can't just delete a page because whatever is in the manual was supported when this version was supported. I think a better option would be to hide all unsupported versions from the version picker. I have created a card for that https://trello.com/c/G23oIBIg/56-hide-all-versions-that-we-dont-support-from-the-version-picker-in-the-operations-manual

Yes, you're right. We discussed it with Ffion.

fiquick:
Thank you for the support @NataliaIvakina and @renetapopova - I'm closing this PR

PR #1938, opened 2024-11-06
Remove deprecation for seedConfig

mnd999:
We are no longer deprecating seedConfig in LTS


NataliaIvakina (approved):
Looks good!
We can safely merge this PR if deprecation has been removed from the code.

PR #1947, opened 2024-11-08
Add a warning about removing disc.service v1

NataliaIvakina:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1947.surge.sh

Updated pages:
- [Cluster server discovery](https://neo4j-docs-operations-1947.surge.sh/operations-manual/5/clustering/setup/discovery)

nick-giles-neo (approved):
And then we can replace "will be removed in a future release." with "has been removed in 2025.01" when that actually goes out, presumably?

NataliaIvakina:
> And then we can replace "will be removed in a future release." with "has been removed in 2025.01" when that actually goes out, presumably?

Yes, exactly. When v1 is actually remove, we'll update the page.

PR #1976, opened 2024-11-21
Update equation for primary node acknowledgement

khoabuiv:
The current equation is (N/2+1). Propose to change to (N/2)+1 to avoid potential confusion for readers.


NataliaIvakina:
Thank you, @khoabuiv for your input! But we decided to keep it as it is. The parenthesis here is a text parenthesis and not a mathematical parenthesis. Even if it were mathematical parenthesis, this change doesn't make sense since multiplication always has presence over addition, and therefore that parenthesis would be redundant.

PR #1980, opened 2024-11-22
add note about database name in query log

LinneaAndersson:
In order to know which database the query is targeting, we do some parsing to find if there is a USE-clause (e.g.  `USE db RETURN 1 AS result`).  If the parsing fails, we are not able to figure out what database the query should be run at and therefor the field in the log will be <none>. 

Please let me know if I can improve the documentation with more details, I just don't want to put information to confuse the user even more.

### DIFF
--- a/modules/ROOT/pages/monitoring/logging.adoc
+++ b/modules/ROOT/pages/monitoring/logging.adoc
@@ -1221,7 +1221,7 @@ Included when xref:configuration/configuration-settings.adoc#config_db.logs.quer
 | Connection details.
 
 | database
-| The database name on which the query is run. This field will be `<none>` if it is not possible to parse and route the query to the correct database.
+| The database name on which the query is run. This field will be `<none>` if the query cannot be parsed and routed to a database.
 
 | executingUser
 | The name of the user executing the query.
--- a/modules/ROOT/pages/monitoring/logging.adoc
+++ b/modules/ROOT/pages/monitoring/logging.adoc
@@ -1221,7 +1221,7 @@ Included when xref:configuration/configuration-settings.adoc#config_db.logs.quer
 | Connection details.
 
 | database
-| The database name on which the query is run.
+| The database name on which the query is run. This field will be `<none>` if it is not possible to parse and route the query to the correct database.
 
 | executingUser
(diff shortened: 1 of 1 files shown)


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1980.surge.sh

Updated pages:
- [Logging](https://neo4j-docs-operations-1980.surge.sh/operations-manual/5/monitoring/logging)

renetapopova:
It looks good, but don't we throw an error as well?

LinneaAndersson:
@renetapopova Yes we do, but there has been confusing about why the database field in the log has been `<none>`. So I just added a tiny comment to the docs :)

modules/ROOT/pages/monitoring/logging.adoc
```diff
@@ -1221,7 +1221,7 @@ Included when xref:configuration/configuration-settings.adoc#config_db.logs.quer
 | Connection details.
 
 | database
-| The database name on which the query is run.
+| The database name on which the query is run. This field will be `<none>` if it is not possible to parse and route the query to the correct database.
```
renetapopova, line 1224:
```suggestion
| The database name on which the query is run. This field will be `<none>` if it is impossible to parse and route the query to the correct database.
```

renetapopova (approved):
A tiny editorial suggestion. Otherwise, looks good.

renetapopova:
@LinneaAndersson, in which version do you want this to be published?

modules/ROOT/pages/monitoring/logging.adoc
LinneaAndersson: For some reason it felt weird to say `impossibe` (don't ask me why 😄). Would it be okay with the change I made to:

"This field will be `<none>` if the query cannot be parsed and routed to a database."

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-1980.surge.sh

Updated pages:
- [Logging](https://neo4j-docs-operations-1980.surge.sh/operations-manual/5/monitoring/logging)

htsedebenham (approved):

PR #2153, opened 2025-02-25
Docs for DEFAULT LANGUAGE

mnd999:
~~Builds on https://github.com/neo4j/docs-operations/pull/2132~~

Depends on https://github.com/neo4j/docs-cypher/pull/1123

Not ready yet

PR #2216, opened 2025-04-02
Add optional SET keyword for create database clauses in Cypher 25

Hunterness:


neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-2216.surge.sh

Updated pages:
- [Create, start, and stop databases](https://neo4j-docs-operations-2216.surge.sh/operations-manual/2025.04/database-administration/standard-databases/create-databases)
- [Database management command syntax](https://neo4j-docs-operations-2216.surge.sh/operations-manual/2025.04/database-administration/syntax)

neo-technology-commit-status-publisher:
View the updated docs at https://neo4j-docs-operations-2216.surge.sh

Updated pages:
- [Create, start, and stop databases](https://neo4j-docs-operations-2216.surge.sh/operations-manual/2025.04/database-administration/standard-databases/create-databases)
- [Database management command syntax](https://neo4j-docs-operations-2216.surge.sh/operations-manual/2025.04/database-administration/syntax)

renetapopova (approved):
Looks good to me.

modules/ROOT/pages/database-administration/syntax.adoc
```diff
@@ -186,10 +262,14 @@ DROP [COMPOSITE] DATABASE name [IF EXISTS] [RESTRICT \| CASCADE ALIAS[ES]] [{DUM
 [[administration-syntax-database-alias-management]]
 == Database alias management command syntax
 
-[options="header", width="100%", cols="1,5a"]
+The following tables cover both local and remote database aliases.
```
Lojjs, line 265: My understanding is that it is only the very next table which cover both, after that it is separate tables for local and remote since there are some variations. So I think you either want to write just "The following table..." instead of "The following tables..." or clarify it a bit.

modules/ROOT/pages/database-administration/syntax.adoc
```diff
@@ -78,7 +78,9 @@ REQUIRE n.propertyName IS NOT NULL
 [[administration-syntax-database-management]]
 == Database management command syntax
 
-The below table covers both standard and composite databases.
+The following tables cover both standard and composite databases.
```
Lojjs, line 81: You might want to clarify that it is the following two (related to SHOW) which cover both, as the third and fourth table after this sentence (related to CREATE) is not for both

Lojjs (changes requested):
The main thing looks fine, just two comments on some wording which might be misleading. I did not test to build it locally, only checked the PR.

modules/ROOT/pages/database-administration/syntax.adoc
Hunterness: I think it's the general `tables` as in all of these tables together covers both? but that sentence came from @renetapopova  so I'll let her decide if we want to update it

modules/ROOT/pages/database-administration/syntax.adoc
Hunterness: same here about it being the tables as a collective that covers both, not specifically each individual one

Hunterness: but maybe something as easy as
```suggestion
The following tables cover standard and composite databases.
```
might do the trick?

Hunterness: (I also think at least DROP covers both, and I guess we'll see what we do for alter composite in https://github.com/neo4j/docs-operations/pull/2153, if we split it or just use the same 🤷)

modules/ROOT/pages/database-administration/syntax.adoc
Lojjs: No I do not think removing 'both' helps. I interpret this sentence as each of the following tables (unclear how many) covers both local and remote alias, which is not true. Some tables cover both, some cover local and some cover remote if I understand the layout of the page correctly

Lojjs: Maybe it is just me that interpret the sentence incorrectly, if you think it is correct already I am fine with leaving it. Same for my other comment which is about the same thing but in another place

Hunterness:
```suggestion
The following sections cover the commands related to local and remote database aliases.
```
is that better?

Hunterness: or feel free to come up with your own suggestion

Hunterness: the sentence isn't supposed to talk about just the next one so changing to `The following table` isn't the way to go

modules/ROOT/pages/database-administration/syntax.adoc
renetapopova: How about:
```suggestion
The database management commands are used to manage standard or composite databases.
```

modules/ROOT/pages/database-administration/syntax.adoc
renetapopova, line 265:
```suggestion
The database alias management commands are used to manage local or remote database aliases.
```

renetapopova (commented):
Added new suggestions.

Lojjs (approved):
Looks good now

PR #2345, opened 2025-05-16
Fix cluster related Docker env variables

NataliaIvakina:


jennyowen:
@NataliaIvakina What's the change? The before and after look identical to me.

NataliaIvakina:
> @NataliaIvakina What's the change? The before and after look identical to me.

@jennyowen, there was only one underscore before the `_address`. Should be two underscores: `__address`

jennyowen (approved):
Good spot, I hadn't noticed. Approved!

PR #2480, opened 2025-07-14
Remove --show-metadata from neo4j-admin backup inspect

renetapopova:
I have removed the `--show-metadata `from the usage and options sections following a failing doc's test and added a few notes on this page explaining the reason. It doesn't look great (I mean to have 6 notes saying the same), because this option is still valid for users on versions before 2025.07 and is also deprecated and not removed. 
I see no other way to document this right now unless we also change the code and make the option visible again.


neo4j-docops-agent:
View the updated docs at https://neo4j-docs-operations-2480.surge.sh

Updated pages:
- [Inspect the metadata of a backup file](https://neo4j-docs-operations-2480.surge.sh/operations-manual/2025.07/backup-restore/inspect)

PR #2512, opened 2025-07-30
Add tabbed examples for multiple node IDs

NataliaIvakina:
The Operations manual covers all releases of 2025.xx series. That's why we cannot simply replace examples with new ones, as new functionality has been added. We also need to retain examples for earlier versions of Neo4j.

PR #2544, opened 2025-08-20
Add tabbed examples for multiple node IDs (#2512)

NataliaIvakina:
The Operations manual covers all releases of 2025.xx series. That's why we cannot simply replace examples with new ones, as new functionality has been added. We also need to retain examples for earlier versions of Neo4j.


neo4j-docops-agent:
View the updated docs at https://neo4j-docs-operations-2544.surge.sh

Updated pages:
- [Import](https://neo4j-docs-operations-2544.surge.sh/operations-manual/2025.07/import)

PR #2611, opened 2025-09-22
docs(browser): add “Clear cached credentials now” how-to

Username46786:
This adds a short step-by-step to the Browser credentials handling page so users can immediately clear cached credentials without waiting for timeouts. It’s based on existing behavior documented on the page (:server disconnect clears credentials; caching can be disabled with browser.retain_connection_credentials=false).

Page: Operations Manual → Security → Browser credentials handling

Rationale: improves task-oriented guidance and reduces user confusion after changing auth settings.


renetapopova (commented):
Hey @Username46786, thanks for your contribution. However, by reading the page, I see that this has already been described. I agree that the structure can be improved, but the content is there. I'll raise an issue to improve the structure. Thanks!

PR #2668, opened 2025-10-20
Update dense node lock description

Roxling:


renetapopova (commented):
I need to open it against the `dev` branch, because `main` is the published version.

PR #2669, opened 2025-10-20
Change http to https

NataliaIvakina:
To fix the failing test on links


NataliaIvakina:
Fails may happen due to the Amazon outage today

PR #2775, opened 2025-12-23
DOCCORE-103 Update NODES promo for NODES AI (#2724)

NataliaIvakina:
Otherwise we have conflicts when merging `dev` into the `publish` branch when publishing the release.


recrwplay (approved):
Whatever makes publishing easier!

(btw, this ad should be removed now anyway)

PR #2064, opened 2025-01-20
Document executionPlanCacheKeyHash and executableQueryCacheHit

renetapopova:
Hi @henriknyman, it seems that those two log entries were missing in the docs. Could you please take a look to see if I have documented them correctly? Also, in which version have they been introduced? Thanks a lot!

PR #2786, opened 2026-01-15
DOCCORE-53:issues relative to executableQueryCacheHit

renetapopova:
This PR hides `executableQueryCacheHit` as it requires the internal setting `internal.dbms.logs.query.query_cache_usage=true` to be enabled in order to be logged in the JSON log. We don't document internal settings.

Related to: https://github.com/neo4j/docs-operations/pull/2064

### DIFF
--- a/modules/ROOT/pages/monitoring/logging.adoc
+++ b/modules/ROOT/pages/monitoring/logging.adoc
@@ -1234,7 +1234,7 @@ The `QueryLogJsonLayout.json` template mimics the 4.x layout and contains the fo
 If the type of the log entry is `query`, these additional fields are available:
 
 .JSON format log entries for log type `query`
-[cols="1m,3a", options="header"]
+[cols="2m,3a", options="header"]
 |===
 | Name
 | Description


modules/ROOT/pages/monitoring/logging.adoc
```diff
@@ -1317,8 +1317,9 @@ Enabled by default only in the JSON format.
 If multiple query executions use the same cached execution plan, their hashes should match.
 For `event=start`, the value is `00000000`.
 
-| executableQueryCacheHit
-| Whether the query string matched a cached execution plan.
+//####Hiding the `executableQueryCacheHit` because it requires the internal setting `internal.dbms.logs.query.query_cache_usage=true` to be enabled in order to be logged.####
+// | executableQueryCacheHit
+// | Whether the query string matched a cached execution plan.
```
henriknyman: I think we should enable this by default in the JSON query log instead, but I guess we can hide it for now if non-Aura users find the absence of this field in their logs confusing

PR #2788, opened 2026-01-15
Prevents the password passed to set-initial-password be saved to command history files and be accessible to other processes

renetapopova:


modules/ROOT/pages/configuration/set-initial-password.adoc
```diff
@@ -24,18 +24,19 @@ Use the xref:configuration/configuration-settings.adoc#config_config_dbms.securi
 .Use the `set-initial-password` command of neo4j-admin
 ====
 
-Set the password for the native `neo4j` user to 'h6u4%krd' before starting the database for the first time.
+You can set the password for the native `neo4j` user before starting the database for the first time. +
+For security reasons, it is recommended to store the password securely and not to include it directly in the command line.
+For example, you can temporarily store the password in a file on disk and then read the file into the command using `cat`, for example:
```
jennyowen: I consulted with @irene221b and I think maybe it would be better to remove lines 29 to 33 entirely.

It's more vague, but avoids us recommending a procedure that isn't so secure. 
There isn't really a nice way of doing the thing. Although one option is to  use a bash user prompt tool such as `dialog` or `whiptail` to prompt for a password and feed it into `neo4j-admin dbms set-initial-password`.

For example, you can do this with whiptail as
`neo4j-admin dbms set-initial-password "$(whiptail --passwordbox "please enter a new neo4j password" 10 50 2>&1 >/dev/tty)"`
or with dialog
`neo4j-admin dbms set-initial-password "$(dialog --passwordbox "please enter a new neo4j password" 10 50 2>&1 >/dev/tty)"`

PR #2806, opened 2026-01-21
Update RBAC limitations page to include vector indexes

Hunterness:
Multiple labels/reltypes/additional properties for vector indexes were introduced in https://github.com/neo4j/docs-cypher/pull/1454.

### DIFF
--- a/modules/ROOT/pages/authentication-authorization/limitations.adoc
+++ b/modules/ROOT/pages/authentication-authorization/limitations.adoc
@@ -52,7 +52,7 @@ In previous versions, they could only be on a single property and a single label
 === Example with denied labels and index on multiple properties
 
 Consider the following full-text example.
-The database has nodes with labels `:User` and `:Person`, and they have properties `name` and `surname`.
+The database contains nodes with the labels `:User` and `:Person`, and the properties `name` and `surname`.
 There are indexes on both properties:
 
 [source, cypher]


modules/ROOT/pages/authentication-authorization/limitations.adoc
```diff
@@ -135,9 +140,9 @@ The steps taken by the query engine would be:
 
 This difference in results is caused by the `OR` relationship between the two properties in the index creation.
 
-=== Denying properties
+=== Example with denied properties on all labels
 
-Now consider denying access on properties, like the `surname` property:
+Continuing the fulltext example above, now consider denying read access on properties for all labels, like the `surname` property:
```
renetapopova: Nobody reads the docs as a book or an article, so therefore, each section is good to be self-contained and not refer to previous examples using "above" and "below", or as in this sentence, "Continuing the fulltext example above".

modules/ROOT/pages/authentication-authorization/limitations.adoc
```diff
@@ -29,22 +29,27 @@ Neo4j lets you create and use indexes to speed up Cypher queries.
...
 This avoids a situation where the same Cypher query would return different results simply because such an index exists.
-The problem is that if you do not know this behavior, you might expect the full-text index to return the same results that a different but semantically similar Cypher query does.
+The problem is that if you do not know this behavior, you might expect the full-text or vector index to return the same results that a different but semantically similar Cypher query does.
 
-=== Example with denied properties
+The examples below all features full-text indexes but also applies to the vector indexes.
+Vector indexes with multiple labels, multiple relationship types or additional properties were introduced in Neo4j 2026.01, before then they could only be on a single property and a single label or relationship type.
 
-Consider the following example.
+=== Example with denied labels and index on multiple properties
```
renetapopova:
```suggestion
[[denied-labels-index-multiple-properties]]
=== Example with denied labels and index on multiple properties
```

modules/ROOT/pages/authentication-authorization/limitations.adoc
renetapopova:
```suggestion
Results from the `surname` property now need to be excluded by the security rules, because they require that the user not to be able to see any `surname` properties.
```

renetapopova:
```suggestion
However, the security model cannot introspect the _Lucene_ query to determine what it will actually do, whether it works only on the allowed `name` property or also on the disallowed `surname` property.
```

modules/ROOT/pages/authentication-authorization/limitations.adoc
```diff
@@ -182,12 +187,61 @@ Therefore, in order to never return results the user should not be able to see,
...
+For that, run the same queries again:
+
+[source, cypher]
+----
+MATCH (n:User)
+WHERE n.name CONTAINS 'ndy'
+RETURN n.name
+----
+
+This query operates similarly to before, but because we do not have the denied traversal on `:Person` the third check on the label is not run.
```
renetapopova: We use "you" to address the user.

renetapopova:
@Hunterness, which version do these changes apply to?

PR #3006, opened 2026-04-23
Document Asynchronous I/O in Neo4j using io_uring

renetapopova:


modules/ROOT/pages/performance/disks-ram-and-other-tips.adoc
```diff
@@ -25,6 +25,53 @@ In this case, database access can have high latencies.
 To achieve maximum performance, it is recommended to provide Neo4j with as much RAM as possible to avoid hitting the disk.
 ====
 
+[role=label--enterprise-edition label--new-2026.04]
+== Asynchronous I/O in Neo4j using io_uring
+
+Neo4j supports asynchronous I/O using the Linux kernel system call interface.
```
MishaDemianenko: using io_uring available on Linux operation system.

`using the Linux kernel system call interface.` - is not really correct since there are multiple ways of doing that on linux and we are using only one of those

modules/ROOT/pages/performance/disks-ram-and-other-tips.adoc
```diff
@@ -25,6 +25,53 @@ In this case, database access can have high latencies.
...
+This enables Neo4j to issue multiple requests concurrently without blocking the main thread and to be notified when the requests are completed.
+For workloads that perform large numbers of I/O operations and do not need to wait for each individual operation to complete, this approach can deliver substantial performance improvements.
+
+This initial delivery enables the background page evictor and checkpointer to take advantage of async I/O when available and configured.
```
MishaDemianenko: `This initial delivery enables the background page evictor and checkpointer to take advantage of async I/O when available and configured.` Should we maybe change this to say that its limited set of components that can utilise that atm and we will expand those abilities over time?
wdyt?

renetapopova: I think it's kind of implicit in this and the next sentence with "This initial delivery" and "While still in its early stages". Product documentation is about the current state anyway, and I didn't want to make any promises. 
Maybe I can change a bit the next sentence to mention that these are not all the components that will be changed.

renetapopova: or maybe better something like:

```suggestion
This initial delivery enables the background page evictor and checkpointer to take advantage of async I/O when available and configured, representing a limited set of components that will expand over time.
```

neo4j-docops-agent:

MishaDemianenko (approved):

PR #3037, opened 2026-05-05
Improve the unquarantine docs

NataliaIvakina:


modules/ROOT/pages/database-administration/standard-databases/errors.adoc
```diff
@@ -191,10 +191,17 @@ The possible values for the optional operation are:
 
 If you choose to clear the current cluster state, the defined database allocation will try to join as a new member.
 This joining process can succeed if and only if a majority of the existing members allow the new member to join.
+
 Assume your database has a topology with three primary allocations.
 If there is only one allocation in `quarantined` mode, then it is safe to choose `replaceStateKeepStore` or `replaceStateReplaceStore`.
```
AnnaSjerling: Maybe we should add "usually" here, since based pn quarantine we cannot 100% guarantee that it will still work. 

"If there is only one allocation in `quarantined` mode, then it is usually safe to choose `replaceStateKeepStore` or `replaceStateReplaceStore`."

modules/ROOT/pages/database-administration/standard-databases/errors.adoc
```diff
@@ -191,10 +191,17 @@ The possible values for the optional operation are:
...
 If there are two allocations in `quarantined` mode (meaning that the database is `quarantined` on two servers), then you should not use concurrently `replaceStateKeepStore` or `replaceStateReplaceStore` for both of them because there would be no majority to let them in.
```
AnnaSjerling: This sentence isn't quite right, I think this sentence more closely describe the situation:

"If there are two allocations in `quarantined` mode (meaning that the database is `quarantined` on two servers), then you should not use `replaceStateKeepStore` or `replaceStateReplaceStore` because there would be no majority to let them in."

AnnaSjerling: I also had another thought here, since we now say we cannot do this in such a situation, should we be pointing to some docs that would guide the user to what should be done in this situation?

I think my preferred approach would be to point to the disaster recovery docs here, which states that a database which has quarantined allocations must be recreated from backup. (server availability section 5)

AnnaSjerling (commented):
I think the new section is very nice! I just have some comments about the other docs that were already there, and I don't think is quite right.

modules/ROOT/pages/database-administration/standard-databases/errors.adoc
```diff
@@ -191,9 +191,18 @@ The possible values for the optional operation are:
...
+
 Assume your database has a topology with three primary allocations.
-If there is only one allocation in `quarantined` mode, then it is safe to choose `replaceStateKeepStore` or `replaceStateReplaceStore`.
-If there are two allocations in `quarantined` mode (meaning that the database is `quarantined` on two servers), then you should not use concurrently `replaceStateKeepStore` or `replaceStateReplaceStore` for both of them because there would be no majority to let them in.
+If there is only one allocation in `quarantined` mode, then it is usually safe to choose `replaceStateKeepStore` or `replaceStateReplaceStore`.
+If there are two allocations in `quarantined` mode (meaning that the database is `quarantined` on two servers), then you should not use `replaceStateKeepStore` or `replaceStateReplaceStore` because there would be no majority to let them in.
+In this case, it is recommended to recreate the quarantined database from a backup.
+See xref:clustering/multi-region-deployment/disaster-recovery.adoc#make-servers-available[Disaster recovery guide -> Make servers available, step 5].
```
AnnaSjerling: Sorry, now I'm being indecisive, but do you think it makes more sense to point it straight to the recreate docs (recreate database, "Use a backup as a seed") that we also point to from the disaster recovery docs? Maybe the disaster recovery docs are a bit harder to understand also because you come into them in the middle of it which is not the recommended approach.

NataliaIvakina: yep, I was thinking the same. I agree it makes more sense to point directly to the recreate procedure page.

NataliaIvakina:
```suggestion
See database-administration/standard-databases/recreate-database.adoc#uri-seed[Recreate a database -> Use a backup as a seed].
```

modules/ROOT/pages/database-administration/standard-databases/errors.adoc
```diff
@@ -191,9 +191,18 @@ The possible values for the optional operation are:
...
 This joining process can succeed if and only if a majority of the existing members allow the new member to join.
+
 Assume your database has a topology with three primary allocations.
-If there is only one allocation in `quarantined` mode, then it is safe to choose `replaceStateKeepStore` or `replaceStateReplaceStore`.
-If there are two allocations in `quarantined` mode (meaning that the database is `quarantined` on two servers), then you should not use concurrently `replaceStateKeepStore` or `replaceStateReplaceStore` for both of them because there would be no majority to let them in.
+If there is only one allocation in `quarantined` mode, then it is usually safe to choose `replaceStateKeepStore` or `replaceStateReplaceStore`.
+If there are two allocations in `quarantined` mode (meaning that the database is `quarantined` on two servers), then you should not use `replaceStateKeepStore` or `replaceStateReplaceStore` because there would be no majority to let them in.
+In this case, it is recommended to recreate the quarantined database from a backup.
+See database-administration/standard-databases/recreate-database.adoc#uri-seed[Recreate a database -> Use a backup as a seed].
```
NataliaIvakina:
```suggestion
See xref:database-administration/standard-databases/recreate-database.adoc#uri-seed[Recreate a database -> Use a backup as a seed].
```

modules/ROOT/pages/database-administration/standard-databases/errors.adoc
```diff
@@ -191,9 +191,18 @@ The possible values for the optional operation are:
 
 If you choose to clear the current cluster state, the defined database allocation will try to join as a new member.
 This joining process can succeed if and only if a majority of the existing members allow the new member to join.
+
 Assume your database has a topology with three primary allocations.
-If there is only one allocation in `quarantined` mode, then it is safe to choose `replaceStateKeepStore` or `replaceStateReplaceStore`.
-If there are two allocations in `quarantined` mode (meaning that the database is `quarantined` on two servers), then you should not use concurrently `replaceStateKeepStore` or `replaceStateReplaceStore` for both of them because there would be no majority to let them in.
+If there is only one allocation in `quarantined` mode, then it is usually safe to choose `replaceStateKeepStore` or `replaceStateReplaceStore`.
+If there are two allocations in `quarantined` mode (meaning that the database is `quarantined` on two servers), then you should not use `replaceStateKeepStore` or `replaceStateReplaceStore` because there would be no majority to let them in.
+In this case, it is recommended to recreate the quarantined database from a backup.
```
AnnaSjerling:
```suggestion
In this case, it is recommended to recreate the database with quarantined allocations from a backup.
```

AnnaSjerling (approved):
I think this looks a lot better now 🎉  thanks for making the PR!

neo4j-docops-agent:

PR #3054, opened 2026-05-08
Update the compatibility table for self-managed Neo4j and Aura versions

NataliaIvakina:


modules/ROOT/pages/database-administration/standard-databases/upload-to-aura.adoc
```diff
@@ -7,16 +7,24 @@
...
+| 5.26 LTS
+| 2026.latest
+
+| 2025.x
+| 2026.latest
 
-| 4.4
-| 4 and 5.latest
+| 2026.x
+| 2026.latest
```
hugofirth, line 27: This might be too minor, but rather than having 2025.x and 2026.x, what if we just said "any >= 2025.01"?

hugofirth (approved):
Minor nit, but approved

modules/ROOT/pages/database-administration/standard-databases/upload-to-aura.adoc
NataliaIvakina: It's a good point but my gut feeling tells me that it's safer to have two separate lines for 2025.x releases and 2026.releases. Later we can combine them into one value, for ex., '2025.01-2026.something LTS'

neo4j-docops-agent:

PR #3062, opened 2026-05-13
Add advance warning on json format for query.log

NataliaIvakina:


modules/ROOT/pages/monitoring/logging.adoc
```diff
@@ -1193,6 +1193,13 @@ In Neo4j Browser and Bloom, the user-provided metadata is always replaced by the
 
 === Use JSON format for the query log
 
+[NOTE]
```
renetapopova: Why do we need this note now? Explaining what will happen in the future. Usually, the docs reflect the current situation, not our plans.

NataliaIvakina: There is a request from PM, see https://linear.app/neo4j/issue/DOCCORE-278/add-advance-warning-in-docs

renetapopova: I see. Maybe then, we can call it **Advanced notice** and make it a warning instead. 
It still feels a bit vague because, as a user, I'll read it, ok, but given that I have no idea when this LTS release will happen, it won't change the current installations, and there is no info that it'll be a breaking change, I'll probably not take it seriously enough.

NataliaIvakina: I'd prefer 'IMPORTANT' rather than WARNING. Warnings are red and can frighten

NataliaIvakina:
```suggestion
[IMPORTANT]
```

renetapopova (approved):

neo4j-docops-agent:

PR #3069, opened 2026-05-14
CCDR: Note remote address resolution is possible

DanGooding:
Document that cross-cluster database replication is possible even with disjoint allocation, where not every upstream server hosts the upstream database. 

We consider this to be the expected default behaviour from customers, so don't want to have too much detail on how or why there is no restriction about upstream allocations, so as not to expose implementation details.

For some more context: we've just implemented remote-address-resolution for replica databases, which (similarly to backups) lets the replica select which upstream servers to pull from. Previously, the replica would fail to create/fail to pull if some upstream server didn't host the upstream database.

Part of [CLUSTER-1367](https://linear.app/neo4j/issue/CLUSTER-1367/remote-address-resolution-for-replicas)


NataliaIvakina (approved):

neo4j-docops-agent:

Issue #2703, opened 2025-11-05
Plugin and Config Docker Mount not documented correctly


EluZet:
As I have mentioned in a [comment](https://github.com/neo4j/docker-neo4j/issues/542#issuecomment-3491015377) on a [docker-neo4j issue](https://github.com/neo4j/docker-neo4j/issues/542) 

The documentation for plugin and docker mount is either linking the wrong folder or not showing that it may be needed to set `server.directories.plugins` to the `/plugins` folder. This is needed since the `/conf` and `/plugins` in the Home Folder have no symlink to the `/plugins` and `/config` folder which docker mounts the volumes on (the Documentation also uses `/config` not `/conf`).

Here again some examples I found, but I'm not sure if it is all:

- [Plugins](https://neo4j.com/docs/operations-manual/current/docker/plugins/)
- [Mounting Volumes](https://neo4j.com/docs/operations-manual/current/docker/mounting-volumes/)
- [Docker Compose](https://neo4j.com/docs/operations-manual/current/docker/docker-compose-standalone/)

For Plugins it works but that is because `server.directories.plugins` is set to `\plugins` instead of `\var\lib\neo4j\plugin` from the beginning. HOWEVER installing plugins through the `NEO4J_PLUGINS=["apoc","apoc-extended"]` does then not work, because `\var\lib\neo4j\plugin` is not set as the Plugin directory and that's where the Docker image installs them.

Both Methods should be using the same folder, or there should be a mention in the documentation that for one of them `server.directories.plugins` needs to be changed.

jennyowen:
The docker entrypoint will always set `NEO4J_server_directories_plugins` config to `/plugins` if it detects that the `/plugins` folder exists, [[source]](https://github.com/neo4j/docker-neo4j/blob/aac3f6c801fecb76df663ab3e9b84067caeea8ea/docker-image-src/calver/coredb/docker-entrypoint.sh#L499). Even if `NEO4J_PLUGINS` is set.

When the entrypoint downloads the plugins, it checks if the `/plugins` folder exists, and downloads plugins there instead of to `/var/lib/neo4j/plugins` IF it exists [[source]](https://github.com/neo4j/docker-neo4j/blob/aac3f6c801fecb76df663ab3e9b84067caeea8ea/docker-image-src/calver/coredb/docker-entrypoint.sh#L146).

[...]

PR #3124, opened 2026-06-10
Fix plugin docker mount documentation

NataliaIvakina:
related to the issue #2703

### DIFF
--- a/modules/ROOT/pages/docker/docker-compose-standalone.adoc
+++ b/modules/ROOT/pages/docker/docker-compose-standalone.adoc
@@ -35,7 +35,8 @@ services:
 ----
 <1> Mount the _/$HOME/neo4j/<..>:_ directories to local directories on your host machine to store logs, configuration, data, and plugins.
 For more information about mounting volumes, see xref:docker/mounting-volumes.adoc[].
-<2> To prevent broken plugin lookups or permission errors, add the strict configuration path mapping directly under the environment variables: `- NEO4J_server_directories_plugins=/plugins`.
+<2> To prevent broken plugin lookups, keep the plugins volume mapping and add the matching configuration path under your environment variables: `- NEO4J_server_directories_plugins=/plugins`.
+The configuration setting does not replace `/plugins` volume mount and only controls where Neo4j looks for plugins.
 <3> Set the `neo4j` username and password.
 
 . Deploy your Neo4j server by running `docker-compose up` from your project folder.
--- a/modules/ROOT/pages/docker/docker-compose-standalone.adoc
+++ b/modules/ROOT/pages/docker/docker-compose-standalone.adoc
@@ -25,9 +25,9 @@ services:
         - /$HOME/neo4j/logs:/logs
         - /$HOME/neo4j/conf:/conf
         - /$HOME/neo4j/data:/data
-        - /$HOME/neo4j/plugins:/plugins
+        - /$HOME/neo4j/plugins:/plugins # <2>
     environment:
-        - NEO4J_AUTH=neo4j/your_password # <2>
+        - NEO4J_AUTH=neo4j/your_password # <3>
     ports:
       - "7474:7474"
       - "7687:7687"
@@ -35,7 +35,8 @@ services:
 ----
 <1> Mount the _/$HOME/neo4j/<..>:_ directories to local directories on your host machine to store logs, configuration, data, and plugins.
 For more information about mounting volumes, see xref:docker/mounting-volumes.adoc[].
-<2> Set the `neo4j` username and password.
+<2> To prevent broken plugin lookups or permission errors, add the strict configuration path mapping directly under the environment variables: `- NEO4J_server_directories_plugins=/plugins`.
+<3> Set the `neo4j` username and password.
 
 . Deploy your Neo4j server by running `docker-compose up` from your project folder.
 +
--- a/modules/ROOT/pages/docker/docker-compose-standalone.adoc
+++ b/modules/ROOT/pages/docker/docker-compose-standalone.adoc
@@ -35,7 +35,7 @@ services:
 ----
 <1> Mount the _/$HOME/neo4j/<..>:_ directories to local directories on your host machine to store logs, configuration, data, and plugins.
 For more information about mounting volumes, see xref:docker/mounting-volumes.adoc[].
-<2> To prevent broken plugin lookups, keep the plugins volume mapping and add the matching configuration path under your environment variables: `- NEO4J_server_directories_plugins=/plugins`.
+<2> If you face broken plugin lookups, keep the plugins volume mapping and add the matching configuration path under your environment variables: `- NEO4J_server_directories_plugins=/plugins`.
 The configuration setting does not replace `/plugins` volume mount and only controls where Neo4j looks for plugins.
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/docker/docker-compose-standalone.adoc
```diff
@@ -23,19 +23,20 @@ services:
...
+        - NEO4J_AUTH=neo4j/your_password # <3>
     ports:
       - "7474:7474"
       - "7687:7687"
     restart: always
 ----
 <1> Mount the _/$HOME/neo4j/<..>:_ directories to local directories on your host machine to store logs, configuration, data, and plugins.
 For more information about mounting volumes, see xref:docker/mounting-volumes.adoc[].
-<2> Set the `neo4j` username and password.
+<2> To prevent broken plugin lookups or permission errors, add the strict configuration path mapping directly under the environment variables: `- NEO4J_server_directories_plugins=/plugins`.
```
NataliaIvakina:
```suggestion
<2> To prevent broken plugin lookups, keep the plugins volume mapping and add the matching configuration path under your environment variables: `- NEO4J_server_directories_plugins=/plugins`.
The configuration setting does not replace `/plugins` volume mount and only controls where Neo4j looks for plugins.
```

modules/ROOT/pages/docker/plugins.adoc
```diff
@@ -59,8 +54,16 @@ graph-data-science
...
 ----
 
-.Install the APOC Core plugin (`apoc`)
-====
+If you rely on the `NEO4J_PLUGINS` environment variable, ensure you map explicit host permissions to avoid container folder ownership blocks and override the server directory mappings inside your environmental configuration:
+
+[source, shell, subs="attributes"]
+----
+# Add to your container environment variables
+NEO4J_server_directories_plugins: /plugins
```
NataliaIvakina:
```suggestion
NEO4J_server_directories_plugins=/plugins
```

modules/ROOT/pages/docker/plugins.adoc
```diff
@@ -59,8 +54,16 @@ graph-data-science
 n10s
 ----
 
-.Install the APOC Core plugin (`apoc`)
-====
+If you rely on the `NEO4J_PLUGINS` environment variable, ensure you map explicit host permissions to avoid container folder ownership blocks and override the server directory mappings inside your environmental configuration:
```
NataliaIvakina:
```suggestion
If you rely on the `NEO4J_PLUGINS` environment variable, override the default server directory mappings inside your environmental configuration:
```

modules/ROOT/pages/docker/docker-compose-standalone.adoc
```diff
@@ -23,19 +23,21 @@ services:
...
+        - NEO4J_AUTH=neo4j/your_password # <3>
     ports:
       - "7474:7474"
       - "7687:7687"
     restart: always
 ----
 <1> Mount the _/$HOME/neo4j/<..>:_ directories to local directories on your host machine to store logs, configuration, data, and plugins.
 For more information about mounting volumes, see xref:docker/mounting-volumes.adoc[].
-<2> Set the `neo4j` username and password.
+<2> To prevent broken plugin lookups, keep the plugins volume mapping and add the matching configuration path under your environment variables: `- NEO4J_server_directories_plugins=/plugins`.
```
NataliaIvakina:
```suggestion
<2> If you face broken plugin lookups, keep the plugins volume mapping and add the matching configuration path under your environment variables: `- NEO4J_server_directories_plugins=/plugins`.
```

NataliaIvakina: I don't want to make it sound like this is a mandatory step

jennyowen: It's not, and also setting the `NEO4J_server_directories_plugins` variable will do nothing for permissions problems.

jennyowen (commented):
I don't think we actually need to do any of these doc changes related to plugins, I think the issue is caused by something else. The change of `/config` to `/conf` is good though we should definitely keep those changes. Maybe put the rest on hold until we can figure out what caused the original issue?

NataliaIvakina:
> I don't think we actually need to do any of these doc changes related to plugins, I think the issue is caused by something else. The change of `/config` to `/conf` is good though we should definitely keep those changes. Maybe put the rest on hold until we can figure out what caused the original issue?

Good idea. I opened another PR to fix the `conf` issue. See PR #3126

modules/ROOT/pages/docker/docker-compose-standalone.adoc
```diff
@@ -25,17 +25,19 @@ services:
...
     ports:
       - "7474:7474"
       - "7687:7687"
     restart: always
 ----
 <1> Mount the _/$HOME/neo4j/<..>:_ directories to local directories on your host machine to store logs, configuration, data, and plugins.
 For more information about mounting volumes, see xref:docker/mounting-volumes.adoc[].
-<2> Set the `neo4j` username and password.
+<2> If you face broken plugin lookups, keep the plugins volume mapping and add the matching configuration path under your environment variables: `- NEO4J_server_directories_plugins=/plugins`.
+The configuration setting does not replace `/plugins` volume mount and only controls where Neo4j looks for plugins.
```
jennyowen: No we don't need this. The original issue is unreproducible and this step is 100% not necessary.

modules/ROOT/pages/docker/plugins.adoc
```diff
@@ -59,8 +54,16 @@ graph-data-science
...
-.Install the APOC Core plugin (`apoc`)
-====
+If you rely on the `NEO4J_PLUGINS` environment variable, override the default server directory mappings inside your environmental configuration:
+
+[source, shell, subs="attributes"]
+----
+# Add to your container environment variables
+NEO4J_server_directories_plugins=/plugins
+----
+
```
jennyowen: again, no this is not necessary. It is done for the user, even when using NEO4J_PLUGINS.

modules/ROOT/pages/docker/mounting-volumes.adoc
```diff
@@ -76,6 +76,10 @@ See xref:docker/security.adoc[]
 See xref:monitoring/metrics/index.adoc[Metrics].
 |===
 
+Keep in mind that mounted directories are connected to Neo4j by matching internal paths.
+For specific directories like `plugins`, you can explicitly override or bind the location by defining the `server.directories.plugins` setting to `/plugins`.
```
NataliaIvakina:
```suggestion
If you build custom plugins directly into the image or want to bypass the Docker entrypoint automation in custom environments, you can explicitly override or bind the location by defining the `server.directories.plugins` setting.
```

NataliaIvakina: @jennyowen , does it make sense to add such a sentence?

jennyowen: yeah it makes sense, although I'd remove the "to `/var/lib/neo4j/plugins`" at the end.

jennyowen (approved):

neo4j-docops-agent:

PR #3171, opened 2026-07-09
Update Query routing decisions page

renetapopova:


ConorNeo (approved):
Nice clear image!

renetapopova:
> Nice clear image!

Thanks for the review, @ConorNeo. Actually, the image is the only thing I didn't change, but looking at it now, I think it needs a bit of updating. It talks about the server leaders instead of databases, and also has just one READ possibility, where there should be two.

ConorNeo:
> > Nice clear image!
> 
> Thanks for the review, @ConorNeo. Actually, the image is the only thing I didn't change, but looking at it now, I think it needs a bit of updating. It talks about the server leaders instead of databases, and also has just one READ possibility, where there should be two.

I think I might be going crazy. I can see no mention of "server leaders" in the image? What is the second read possibility of which you talk? :)

renetapopova:
> > > Nice clear image!
> > 
> > 
> > Thanks for the review, @ConorNeo. Actually, the image is the only thing I didn't change, but looking at it now, I think it needs a bit of updating. It talks about the server leaders instead of databases, and also has just one READ possibility, where there should be two.
> 
> I think I might be going crazy. I can see no mention of "server leaders" in the image? What is the second read possibility of which you talk? :)

I mean, we say "Is the local member the leader for that database?" but servers can't be leaders anymore; one of the databases is the leader for the rest of the same databases. 

> What is the second read possibility of which you talk? :)

Screenshot 2026-07-10 at 14 51 08

renetapopova:
Hey @ConorNeo, could you please review the diagram? I have updated it according to the described steps, and if you are happy with it, I'll ask Lidia to reproduce it in the Neo4j style and colours.

modules/ROOT/images/routing-decisions.svg
NataliaIvakina, line 1: If we want to update the image, should we add also 'Target database is a sharded database' to the step 3?

renetapopova: I am not sure I understand. There are no sharded databases here. Do you mean to also add what the query routing decisions would be if the target database is an SPD?

renetapopova: Maybe @ConorNeo can chime in here. I have no idea what the differences would be. We'll also need to update the steps.

NataliaIvakina: > Do you mean to also add what the query routing decisions would be if the target database is an SPD?

 yes, that's what I meant 😄 To add information about query routing in an SPD. But only if we want it. This is just a question.

modules/ROOT/pages/database-administration/routing-decisions.adoc
```diff
@@ -4,51 +4,50 @@
...
-* Execute the query in the open transaction.
-
-== Illustrated routing decision tree
-image::routing-decisions.svg[width=800, title="Illustrated routing decision tree"]
+_Internal_::
+** If the URI scheme is `bolt://` (routing disabled), the location is _Local_.
+** If the transaction mode is `READ` and the database is hosted on _this server_, the location is _Local_.
+** If the transaction mode is `READ` and the database is hosted on _another server_ with
+xref:clustering/setup/routing.adoc#clustering-routing[server-side routing] xref:configuration/configuration-settings.adoc#config_dbms.routing.enabled[enabled], the location is _Remote_ (using the xref:configuration/configuration-settings.adoc#config_server.routing.advertised_address[routing advertised address] of that server).
+If server-side routing _is not_ enabled, then fail.
```
NataliaIvakina:
```suggestion
If server-side routing _is not_ enabled, then the operation fails.
```

modules/ROOT/images/routing-decisions.svg
ConorNeo: I don't think it matters here, I think that remote alias and or internal to the DBMS covers it, nothing special happens for SPD, just routes to it fine

ConorNeo (approved):
Discussion took place on slack looks good!

renetapopova:
Hey @lidiazuin, the diagram is ready to be refreshed in the Neo4j style.

neo4j-docops-agent:

PR #3189, opened 2026-07-16
Update the deprecation tables to be ordered by deprecation version

renetapopova:


NataliaIvakina (approved):
I really like the new table structure. Thank you, Reneta.
My only concern is about using Replacement instead of Comment. In some tables, it looks much better, I agree. But Comment as a header is broader than the Replacement, and configs, metrics, procedures are not always replaced, sometimes they are just removed (will be removed). Also, from time to time it's good to add some other details not related to the replacement/removal.

Can we still use Comment in the Metrics table?
If the Replacement cell is empty, does it always mean that there is no replacement, and this item will be removed?

renetapopova:
> I really like the new table structure. Thank you, Reneta. My only concern is about using Replacement instead of Comment. In some tables, it looks much better, I agree. But Comment as a header is broader than the Replacement, and configs, metrics, procedures are not always replaced, sometimes they are just removed (will be removed). Also, from time to time it's good to add some other details not related to the replacement/removal.
> 
> Can we still use Comment in the Metrics table? If the Replacement cell is empty, does it always mean that there is no replacement, and this item will be removed?

About your second question, I would guess so. I haven't added or removed any information. Where there was no replacement or other information, I just didn't add anything.

neo4j-docops-agent:

PR #3198, opened 2026-07-21
Update the version for running a WHERE clause on the system database

renetapopova:


modules/ROOT/pages/procedures/call-procedures.adoc
```diff
@@ -347,7 +347,7 @@ CALL dbms.unquarantineDatabase(server, database)
 
 Nothing is returned from the query, as it ends with a procedure call of a void procedure.
 
-[role=label--new-2026.06]
+[role=label--new-2026.07 label--cypher-25]
```
Hunterness: not sure it's actually restricted to Cypher 25, it's not distinguished at a parser level (both parsers already parses the query, and then we were just blocking it when building the execution plan for the procedure when run against system, which we have now removed. But I don't think did any checks for cypher version on that.

Hunterness: Update: I did try it locally in a test and it also works in Cypher 5

```suggestion
[role=label--new-2026.07]
```

Hunterness (approved):

neo4j-docops-agent:

PR #3177, opened 2026-07-14
Revert "Add backup option `--skip-producing-empty-diff`"

DanGooding:
Reverts neo4j/docs-operations#3162

See:
- https://github.com/neo-technology/neo4j/pull/38844
- https://github.com/neo4j/docs-operations/pull/3177 <- I think tests will fail until this gets merged to remove the option in dev

PR #3162, opened 2026-07-07
Add backup option `--skip-producing-empty-diff`

DanGooding:
Document a new public option for `backup` that disables the creation of empty diff backups (containing zero transactions).

This option is incompatible with `--include-metadata` so we require `--include-metadata=none when it is specified.

The option is added in
- https://github.com/neo-technology/neo4j/pull/38434

CLUSTER-1675

PR #3209, opened 2026-07-27
Add backup option `--skip-empty-diffs`

DanGooding:
Documents the `neo4j-admin database backup --skip-empty-diffs` option, which is being made public in https://github.com/neo-technology/neo4j/pull/39121 (part of CLUSTER-1675).

This option was previously documented as `--skip-producing-empty-diff` in #3162, but that was reverted in #3177 when the flag was hidden again in the product. It is now public under the shorter name `--skip-empty-diffs`, so this re-adds it with the new name.

Changes:
- Add the `--skip-empty-diffs[=true|false]` row to the options table, labelled `Introduced in 2026.08`.
- Refresh the command synopsis to match the current CLI output.

Both the synopsis and the option description are taken verbatim from the updated `OnlineBackupCommandTest.printUsageHelp` expectation in that PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Part of CLUSTER-1675


NataliaIvakina (approved):
Looks good!
I'll keep an eye on this PR and merge it after publishing 2026.07

neo4j-docops-agent:

PR #3210, opened 2026-07-28
update docs around non-obfuscated literals

fastlag:
with the change of no longer obfuscating some literals (SKIP/LIMIT & FIELDTERMINATOR), the docs should reflect that as well.

### DIFF
--- a/modules/ROOT/pages/configuration/configuration-settings.adoc
+++ b/modules/ROOT/pages/configuration/configuration-settings.adoc
@@ -2926,7 +2926,7 @@ m|+++false+++
 [frame="topbot", stripes=odd, grid="cols", cols="<1s,<4"]
 |===
 |Description
-a|If true, obfuscates all literals in a query before writing the query to the query log. Note that node labels, relationship types, skip/limit values, field terminators in load csv, and map property keys, while being literals, remain visible. Changing the setting will not affect queries that are cached. To apply the change immediately, you must also call `CALL db.clearQueryCaches()`. It is recommended to set this setting to `true` in production.
+a|If true, obfuscates all literals in a query before writing the query to the query log. Note that node labels, relationship types, skip/limit values, field terminators in `LOAD CSV`, and map property keys, while being literals, remain visible. Changing the setting will not affect cached queries. To apply the change immediately, you must also run `CALL db.clearQueryCaches()`. It is recommended to set this setting to `true` in production.
 |Valid values
 a|A boolean.
 |Default value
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/configuration/configuration-settings.adoc
```diff
@@ -2926,7 +2926,7 @@ m|+++false+++
 [frame="topbot", stripes=odd, grid="cols", cols="<1s,<4"]
 |===
 |Description
-a|If true, obfuscates all literals in a query before writing the query to the query log. Note that node labels, relationship types, and map property keys remain visible. Changing the setting will not affect queries that are cached. To apply the change immediately, you must also call `CALL db.clearQueryCaches()`. It is recommended to set this setting to `true` in production.
+a|If true, obfuscates all literals in a query before writing the query to the query log. Note that node labels, relationship types, skip/limit values, field terminators in load csv, and map property keys, while being literals, remain visible. Changing the setting will not affect queries that are cached. To apply the change immediately, you must also call `CALL db.clearQueryCaches()`. It is recommended to set this setting to `true` in production.
```
renetapopova: Because I see that the description still hasn't been updated in the monorepo, could we update it a bit to:
```suggestion
a|If true, obfuscates all literals in a query before writing the query to the query log. Note that node labels, relationship types, skip/limit values, field terminators in `LOAD CSV`, and map property keys, while being literals, remain visible. Changing the setting will not affect cached queries. To apply the change immediately, you must also run `CALL db.clearQueryCaches()`. It is recommended to set this setting to `true` in production.
```

renetapopova (approved):
Looks very good — just minor suggestions, if it is still possible to update the description in the monorepo. They should be the same.

fastlag:
> Looks very good — just minor suggestions, if it is still possible to update the description in the monorepo. They should be the same.

Thank you , I applied the suggestion (looks a lot better). I would like to update the monorepo doc as well, but from what I have understood that is quite a long process and I go on vacation tomorrow. I'll create a ticket for someone in the team to do it when they have time!

renetapopova:
> > Looks very good — just minor suggestions, if it is still possible to update the description in the monorepo. They should be the same.
> 
> Thank you , I applied the suggestion (looks a lot better). I would like to update the monorepo doc as well, but from what I have understood that is quite a long process and I go on vacation tomorrow. I'll create a ticket for someone in the team to do it when they have time!

Why is it a long process to update a description?

neo4j-docops-agent:

PR #3214, opened 2026-07-30
Backup Cosi local storage requirements

yirensum:
This PR is a result of this conversation with [field](https://neo4j.slack.com/archives/C03CEJCAS/p1783991250177489) about storage requirements needed when doing backups via COSI.

Part of CLUSTER-1876

### DIFF
--- a/modules/ROOT/pages/backup-restore/online-backup.adoc
+++ b/modules/ROOT/pages/backup-restore/online-backup.adoc
@@ -210,22 +210,14 @@ If you want to force a full backup, use `FULL`.
 ====
 The `--to-path=<path>` option can also back up databases into AWS S3 buckets, Google Cloud storage buckets, and Azure buckets.
 For more information, see <<online-backup-cloud-storage>>.
-====
 
-[NOTE]
-====
-The `--temp-path` option can address potential issues related to disk space when performing backup-related commands, especially when cloud storage is involved.
+Even when `--to-path` points to a cloud storage bucket, the backup process requires a temporary local directory to store the database store files and transaction logs before creating and streaming the backup artifact to the cloud destination.
+The temporary directory must therefore have enough free space to hold the entire backup.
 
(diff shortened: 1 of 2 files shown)


NataliaIvakina:
The test failure is covered in https://github.com/neo4j/docs-operations/pull/3209.
Do not rebase this PR to make the test green! The new option has to be published with 2026.08, while this addition can be published anytime

modules/ROOT/pages/backup-restore/online-backup.adoc
NataliaIvakina: I think it would be better to rewrite two notes because we're repeating the same things using different wording.

I'm suggesting replacing lines 209-229 with the following single note

NataliaIvakina: My suggestion is the following:

````
====
The `--to-path=<path>` option can also back up databases into AWS S3 buckets, Google Cloud storage buckets, and Azure buckets.
For more information, see <<online-backup-cloud-storage>>.

Even when `--to-path` points to a cloud storage bucket, the backup process requires a temporary local directory to store the database store files and transaction logs before creating and streaming the backup artifact to the cloud destination.
The temporary directory must therefore have enough free space to hold the entire backup.

If `--to-path` is not specified or points to a cloud storage bucket, Neo4j creates the temporary directory in its current working directory.
If that location does not have sufficient free disk space, the backup can fail.

To avoid this issue, it is strongly recommended to specify `--temp-path` and point it to a local directory with sufficient free disk space, especially when backing up to cloud storage.
====
````

yirensum: I think that looks better as well, added in [14a5806](https://github.com/neo4j/docs-operations/pull/3214/commits/14a58066ea87760da9b942f96063300c45860515)

NataliaIvakina (approved):
Looks good!

neo4j-docops-agent:

PR #1891, opened 2024-10-21
Add info about the `--temp-path` option

NataliaIvakina:

PR #3220, opened 2026-07-31
Clarify backup COSI local storage requirements

NataliaIvakina:
See the original PR #3214

### DIFF
--- a/modules/ROOT/pages/backup-restore/online-backup.adoc
+++ b/modules/ROOT/pages/backup-restore/online-backup.adoc
@@ -188,22 +188,21 @@ If you want to force a full backup, use `FULL`.
 |
 |===
 
+[[temp-path-explanation]]
 [NOTE]
 ====
 The `--to-path=<path>` option can also back up databases into AWS S3 buckets (from Neo4j 5.19), Google Cloud storage buckets (from Neo4j 5.21), and Azure buckets (from 5.24).
 For more information, see <<online-backup-cloud-storage>>.
-====
 
-[NOTE]
-====
 Neo4j 5.24 introduces the `--temp-path` option to address potential issues related to disk space when performing backup-related commands, especially when cloud storage is involved.
 
-If `--temp-path` is not set, a temporary directory is created inside the directory specified by the `--path` option.
+When `--to-path` points to a cloud storage bucket, the backup process still requires a temporary local directory to store the database store files and transaction logs before creating and streaming the backup artifact to the cloud destination.
+The temporary directory must therefore have enough free space to hold the entire backup.
 
-If you don't provide the `--path` option or if your provided path points to a cloud storage bucket, a temporary folder is created inside the current working directory for Neo4j.
-This fallback option can cause issues because the local filesystem (or the partition where Neo4j is installed) may not have enough free disk to accommodate the intermediate computation.
+If `--to-path` is not specified or points to a cloud storage bucket, Neo4j creates the temporary directory in its current working directory.
+If that location does not have sufficient free disk space, the backup can fail.
 
-Therefore, it is strongly recommended to provide a `--temp-path` option when executing a backup-related command, especially if the folder provided in the `--path` option points to a cloud storage bucket.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/backup-restore/online-backup.adoc
```diff
@@ -188,22 +188,21 @@ If you want to force a full backup, use `FULL`.
...
-====
 Neo4j 5.24 introduces the `--temp-path` option to address potential issues related to disk space when performing backup-related commands, especially when cloud storage is involved.
 
-If `--temp-path` is not set, a temporary directory is created inside the directory specified by the `--path` option.
+When `--to-path` points to a cloud storage bucket, the backup process still requires a temporary local directory to store the database store files and transaction logs before creating and streaming the backup artifact to the cloud destination.
+The temporary directory must therefore have enough free space to hold the entire backup.
 
-If you don't provide the `--path` option or if your provided path points to a cloud storage bucket, a temporary folder is created inside the current working directory for Neo4j.
-This fallback option can cause issues because the local filesystem (or the partition where Neo4j is installed) may not have enough free disk to accommodate the intermediate computation.
+If `--to-path` is not specified or points to a cloud storage bucket, Neo4j creates the temporary directory in its current working directory.
```
yirensum: Apologies, I realized that I misread this yesterday, and that it's possibly incorrect (and needs correcting in dev as well).

In my head it should be something like:

``` 
The temporary path can be specified by "--temp-path".
If "--temp-path" is unspecified, Neo4j creates the temporary directory in it's current working directory. 
```

What this means is that cloud storage bucket or not, the user can specify a temp-path where everything is downloaded to - does this sort of make sense?

NataliaIvakina: No probs, happy to fix. But I'd like to double check because now I'm confused.

So, `neo4j-admin database backup` has two options `to-path` and `temp-path`.  The `temp-path` was introduced only in 5.24.

We stated earlier that if you don't specify `--path` (I think we meant `to-path`, see PR https://github.com/neo4j/docs-operations/pull/1891, we updated a few pages there) or if it points to a cloud bucket, the temporary dir is created in the current working dir.
Then, `temp-path` was introduced. If you don't specify both `temp-path` and `to-path`, or if you don't specify `temp-path` and `to-path` points to cloud bucket, then the temp dir is created in the current Neo4j's working dir. Is this correct or not?

yirensum: so if we don't specify `--to-path`, it'll error out, we should always specify a `--to-path`, I tested this locally:
```
 ./bin/neo4j-admin database backup --prefer-diff-as-parent --type=FULL neo4j 
-----------------------------------------------------
Command Failed
Missing option: '--to-path=<path>'.
-----------------------------------------------------
```

As such the combinations should as follows:
```
--to-path=s3://bucket      --temp-path=null                      -> temp-path where neo4j-admin is
--to-path=s3://bucket      --temp-path=local/temp/         -> temp-path in local/temp/
--to-path=/local/backup/   --temp-path=null                    -> temp-path in /local/backup/
--to-path=/local/backup/   --temp-path=local/temp/       -> temp-path in local/temp/
```

yirensum: so before 5.24, I'm guessing it follows the `--temp-path=null` logic. However, sometimes customers have multiple-disks attached, and want to use another location with more space for the download (hence why we introduced `--temp-path` so they can specify that)

NataliaIvakina:
```suggestion
The temporary path can be specified by the `--temp-path` option.
If `--temp-path` is unspecified, Neo4j creates the temporary directory in its current working directory. 
```

neo4j-docops-agent:

modules/ROOT/pages/backup-restore/online-backup.adoc
NataliaIvakina: I also raised a PR https://github.com/neo4j/docs-operations/pull/3222 in `dev`

